Issue #265, opened 2014-10-02
Problem Content


edwoodward:
This is a list of links to content that is returned in search results, but gives a 500 error when viewed.
- Stereo Depth Map Construction
  - http://cnx.org/contents/ff4baa7b-da6e-4852-ab11-404ff8811679@1.1
- EE Textbook
  - http://cnx.org/contents/c992c08b-2130-4e61-96c8-74d4c3092cc8@1.1/EE_Textbook

reedstrm:
Hmm, I wonder how  they managed to convince the existing system to put a collection inside a collection: as Michael mentioned on the PR, this is not supposed to be possible.

reedstrm:
That legacy links shows what happens to a collection that has the wrong status/state set, somehow. That's the collection editor interface.

edwoodward:
It has happened before but I don't remember how. Copy and paste comes to 
mind, but I don't think that is possible unless the old collection 
editor allowed it.

reedstrm:
This is from 2009, and Mark Horner's the author, so various forms of hackery could have been in play :-)

edwoodward:
This appears to be fixed. Closing

PR #277, opened 2015-01-07
Choose the first node in trees by documentid and null parent_id

karenc:
This fixes an issue with generating the collection tree json for a
collection that is also in another collection.

See #265


karenc:
This is a change to the `tree_to_json` sql function, it will require a manual update to replace the function stored in the database.

pumazi:
I'm fairly sure a collection within a collection is invalid. Apparently they exist though.

Perhaps we need to migrate/fix these collections rather than allowing them to exist? (cc @reedstrm )

PR #280, opened 2015-01-23
Publish to legacy

karenc:
This is necessary for https://github.com/Rhaptos/Products.RhaptosModuleStorage/pull/7 to work


karenc:
cnx-upgrade to_html uses `cnxarchive.to_html`, should we update cnx-upgrade too?

PR #284, opened 2015-01-27
Aux author extras

### LINKED ISSUE #194: Auxiliary author(s) are not returned if there are no results
`results.auxiliary.authors` returns an empty array if the result set is also empty.

Local Example: http://localhost/search?q=author:collegi%20authorID:e5a07af6-09b9-4b74-aa7a-b7510bee90b8

Live Examples:
http://cnx.org/search?q=author:Jos%C3%A9%20authorID:anazagasty
http://cnx.org/search?q=author:Pose%20authorID:anazagasty

It also appears to be a bug that the first live example returns no results, since http://cnx.org/search?q=author:Jos%20authorID:anazagasty works (appears to be an issue with decoding the url value for `é`).

rich-hart:
In the case where a search is performed with an authorId as a filter,
it is possible for the database to return no results even if the author
exists in the database. Therefore, the database is queried a second
time for contact information associated with only the authorIds. The
author information is then used to update the results returned by the
first database query.

Fix #194

### DIFF
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -52,6 +52,7 @@ SQL = {
     'get-module-versions': _read_sql_file('get-module-versions'),
     'get-subject-list': _read_sql_file('get-subject-list'),
     'get-featured-links': _read_sql_file('get-featured-links'),
+    'get-users-by-ids': _read_sql_file('get-users-by-ids'),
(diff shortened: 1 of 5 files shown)


karenc:
Looks perfect

PR #286, opened 2015-02-04
Adding an index field to results when an aux author special case search is performed

rich-hart:
Adding this index field to the results json data structure allows Webview to properly render the authorID strings inside search filter icons.


pumazi:
The logic looks fine. However, I'm seeing spacing issues again.

rich-hart:
whoops! forgot to run a pep8 check.  thnx!

pumazi:
I took care of the spacing, no worries.

PR #287, opened 2015-02-11
Fix tests to only check for general conversion

pumazi:
The generalized check allows us to change rhaptos.cnxmlutils
without worrying that these tests will fail.


pumazi:
BTW, master is failing. That is why these changes were created.

PR #289, opened 2015-02-12
Adding search api documenation

rich-hart:


edwoodward:
This looks good.

pumazi:
Please adjust the person-info values to not include: website, suffix and email(s). This won't reflect the current state of things, but I'd rather not make the adjustment later.

search_api_doc.rst
```diff
@@ -0,0 +1,225 @@
...
+Parameters:
+
++---------+---------+---------------------------------+
+| Field   | Type    | Description                     |
++---------+---------+---------------------------------+
+| QUERY   | String  | General search string           |
++---------+---------+---------------------------------+
+| AUTHOR  |         | Search by author contact info.  |
++---------+---------+---------------------------------+
+| ID      |         | Search by a specific author ID  |
```
pumazi: This should be authorID and the case of these should match the case we parse for. (I seem to recall a conversation that happened long long about involving authorId vs authorID. I'd hate to rehash that conversations because someone read this wrong. =P

search_api_doc.rst
```diff
@@ -0,0 +1,225 @@
...
+
+Brakets denote an optional parameter.
+
+
+Parameters:
+
++---------+---------+---------------------------------+
+| Field   | Type    | Description                     |
++---------+---------+---------------------------------+
+| QUERY   | String  | General search string           |
```
pumazi: ``q`` rather than QUERY shouldn't be in the values you are listing here. It is a request parameter rather than a search term/limit parameter.

pumazi:
I don't see all the available limits in this.

search_api_doc.rst
rich-hart: So this was something that I was going to bring up for later.   I think what your saying is that the fields column in the table should reflect the case sensitive filter string that we parse for.  For example, our API will accept 'authorID:OpenStaxCollege' but not 'AUTHORID:OpenStaxCollege'.  Is that right?  

If that's the case, you'll notice in the line:
```
QUERY [authorID:ID] 
```
the lower case values such as 'authorID' denote the key needed for the API to interpret the request and the the uppercase values such as 'ID' denote the variables that the user specifies.   I know that its somewhat backwards because we are used to denoting constant values link 'authorID' as fully UPPERCASE strings.  But our API flag strings are set at the moment, so that's why the table field values are all capital.

E.g. If 
```
QUERY = Physics
ID = OpenStaxCollege
```

then 

```
QUERY [authorID:ID]  -->  Physics authorID:OpenStaxCollege
```

rich-hart:
What other limits am I missing?

pumazi:
Off the top of my head... licensor, maintainer, parentauthor, etc.

They should be listed in the search module. Some keen eye balls may be required.

search_api_doc.rst
pumazi: Ah ok, I see what you are saying. Looks fine, now that I understand where you were going with it.

pumazi:
This looks good. @reedstrm does this meet your expectations?

PR #290, opened 2015-02-16
Replace FDW to accounts with a users table

pumazi:
See also [trello card](https://trello.com/c/2Z2OEcSq/1116-convert-users-table-from-fdw-to-shadow-table)


reedstrm:
This appears to do what it says on the tin: removes the fdw and associated machinery.

Issue #292, opened 2015-02-23
Print styles not displaying in metadata


amberwebb:
After publishing, the print styles selection is not displaying in metadata.

pumazi:
See also, Connexions/cnx-publishing#78

pumazi:
Now in master from #296

PR #294, opened 2015-02-25
adding print_style column to modules and latest_modules tables

rich-hart:
Database changes to add support for issue #292


rich-hart:
Overalls has been having trouble with this pull request.  The ``` continuous-integration/travis-ci/pr ``` test originally returned 'OK' but not it has spontaneously changed to fail.  Any ideas on the reason?

PR #295, opened 2015-02-25
making whitespace changes to sql files

rich-hart:
I'm finding it hard to read these SQL files because there is not enough indention / separation of column names and nested sql statements.  So I've run these scripts through formatting programs followed by some manual modifications.  These formatting changes are by no means perfect but will be helpful in the short term for modifying database queries.   

The file modifications were made with two separate formatting programs: ```get-module-metadata.sql``` was formatted with the online web service http://www.dpriver.com/pp/sqlformat.htm and ```get-module.sql``` was modified with a python module ```sqlparse```.

### DIFF
--- a/cnxarchive/sql/get-module-metadata.sql
+++ b/cnxarchive/sql/get-module-metadata.sql
@@ -1,100 +1,181 @@
--- ###
--- Copyright (c) 2013, Rice University
--- This software is subject to the provisions of the GNU Affero General
--- Public License version 3 (AGPLv3).
--- See LICENCE.txt for details.
--- ###
-
--- arguments: id:string; version:string
-SELECT row_to_json(combined_rows) AS module
-FROM (SELECT
-  m.uuid AS id,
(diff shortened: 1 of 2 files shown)


pumazi:
Is there an **open source** command line version of this service/tool?

rich-hart:
The ```sqlparse``` python module is open source but its parameters are vary limited.  It's source code is on https://github.com/andialbrecht/sqlparse and its documentation says that it has a command line tool.

rich-hart:
The nice thing about ```sqlparse``` is that it does have a parameter for printing SQL code as python readable strings.

pumazi:
Ok. Show me what that looks like. I'm not opposed to changing repr(-pretty) formats as long as I can reproduce it with an open source tool.

rich-hart:
I used ```sqlparse``` on one of the sections of cnx-publishing ( https://gist.github.com/rich-hart/77bf6f3123a87a216c70 ). The output isn't great unfortunately.

PR #297, opened 2015-03-04
extend exports info with dates and sizes

reedstrm:


pumazi:
Looks like legit failures. You'll need to patch (via ``mock.patch``) the datetime outputs.

reedstrm:
arrrgh! timezones!

.travis.yml
```diff
@@ -34,6 +34,7 @@ before_install:
   - sudo /usr/bin/pip install coveralls
 install:
   - sudo /usr/bin/python setup.py install
+  - sudo /usr/bin/pip install mock
```
pumazi: Include in ``setup.py:setup(.., test_require={}, ..)``. See also, https://github.com/Connexions/cnx-publishing/blob/908221cce57e8c373d78acf51c437b1806317ec4/setup.py#L27

cnxarchive/tests/test_views.py
```diff
@@ -989,6 +994,7 @@ def test_exports_without_version(self):
         self.assertEqual(cm.exception.headers,
                          [('Location', '/exports/{}@5.pdf'.format(id))])
 
+    @mock.patch('cnxarchive.utils.fromtimestamp', mock.Mock(side_effect=testing.mocked_fromtimestamp))
```
karenc: This works when I try to run the tests individually.  It fails when I'm running all the tests in ViewsTestCase.

Doing this instead works locally for me: `@mock.patch('cnxarchive.views.fromtimestamp', mock.Mock(side_effect=testing.mocked_fromtimestamp))` cnxarchive.**views**.fromtimestamp instead of cnxarchive.utils.fromtimestamp.

May be it's because we import just the function `fromtimestamp` into `cnxarchive.views` from `cnxarchive.utils` instead of importing `cnxarchive.utils` and doing `cnxarchive.utils.fromtimestamp`.

cnxarchive/tests/testing.py
```diff
@@ -27,6 +30,8 @@
 here = os.path.abspath(os.path.dirname(__file__))
 config_uri = None
 
+def mocked_fromtimestamp(timestamp):
```
karenc: Should we add a comment here like: `# Always return 2015-03-04 10:03:29-08:00`?

cnxarchive/views.py
```diff
@@ -120,18 +120,21 @@ def get_export_allowable_types(cursor, exports_dirs, id, version):
...
+                'state': state,
                 'details': type_info['description'],
                 'path': u'/exports/{}@{}.{}/{}'.format(id, version, type_name,
                                                       filename),
                 }
         except ExportError as e:
-            # file not found, so don't include it
-            pass
+            # Some other problem, skip it
+                pass
```
karenc: pass shouldn't be indented

PR #299, opened 2015-03-06
Include the legacy persons table in the database schema.

pumazi:


reedstrm:
This will make generating schema diffs easier

PR #302, opened 2015-03-11
Add a moderation boolean to the user entries.

pumazi:
This will allow publishing to know that a user has been vetted
for publication. If not, the publication will be put into a
wait state where an moderator will need to moderate the content.

### DIFF
--- a/cnxarchive/sql/schema/schema.sql
+++ b/cnxarchive/sql/schema/schema.sql
@@ -617,7 +617,9 @@ CREATE TABLE users (
   last_name TEXT,
   full_name TEXT,
   suffix TEXT,
-  title TEXT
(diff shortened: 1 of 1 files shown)


pumazi:
```sql
ALTER TABLE users
    ADD COLUMN is_moderated boolean;

UPDATE users SET (is_moderated) = ('t')
    WHERE username IN (
        SELECT DISTINCT unnest(authors) FROM modules
        UNION
        SELECT DISTINCT unnest(maintainers) FROM modules
        UNION
        SELECT DISTINCT unnest(personids) FROM moduleoptionalroles);
```

reedstrm:
Some reason ypu left licensors out of the "preapproved" list?

PR #306, opened 2015-03-26
Add the moderation flag to the test data user entries

pumazi:


reedstrm:
This may be a bit late, but I'm not excited about the name is_moderated: it's not clear to me at first glance if that's a user that needs moderation, or has 'passed' moderation. has_published is the actual policy, but it's not the mechanism. 'vetted' or 'approved' seems to be the actual sense. 'no_training_wheels' 'probably_not_a_spammer' ' big_boy/girl' 'not_a_bot' ...

PR #310, opened 2015-04-10
Modifying app_parser, app_settings, and initdb functions

### LINKED ISSUE #304: initdb has SQL permissions issue
Related to issue https://github.com/Connexions/cnx-publishing/issues/49

rich-hart:
This allows for a superuser to be provided to the 'initdb' function so that the 'plpythonu' language and methods can be created without permission errors.  After which, a non-super user can be used to complete the rest of the install.   

Changes also include a new test class which makes one call to 'initdb' to be made during the unit test class setup.  Each test is cleaned up by rolling back the changes to the database with the completion of each test function. 

Fixes #304

### DIFF
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -36,6 +36,10 @@ DB_SCHEMA_FILES = (
 DB_SCHEMA_FILE_PATHS = tuple([os.path.join(DB_SCHEMA_DIRECTORY, dsf)
                               for dsf in DB_SCHEMA_FILES])
 
+EXAMPLE_DATA_FILEPATHS = (
+    config.TEST_DATA_SQL_FILE,
+)
+
 
 def _read_sql_file(name):
     path = os.path.join(SQL_DIRECTORY, '{}.sql'.format(name))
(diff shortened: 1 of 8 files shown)


cnxarchive/database.py
```diff
@@ -60,19 +64,33 @@ def _read_sql_file(name):
...
+    else:
+        connection_string = settings['db-connection-string']
     with psycopg2.connect(connection_string) as db_connection:
         with db_connection.cursor() as cursor:
             for schema_filepath in DB_SCHEMA_FILE_PATHS:
                 with open(schema_filepath, 'r') as f:
                     cursor.execute(f.read())
+            # allow standard user to access the  'plpythonu' langauge.
+            cursor.execute(
+                "ALTER LANGUAGE plpythonu OWNER TO {user}".format(**settings))
```
karenc: It's probably not critical here but shouldn't we always use query placeholders like `cursor.execute('ALTER LANGUAGE plpythonu OWNER TO %(user)s', settings)`?

rich-hart: That's actually the way that I tried to implement the SQL the first time.  The problem is that when you use

```python 
cursor.execute('ALTER LANGUAGE plpythonu OWNER TO %(user)s', settings)
```
the psycopg library interprets users to be a string rather than a postgres identifier.  For example, if in python you set ' settings['user'] = 'karen' then resulting SQL command using the psycopg library would be
```SQL
ALTER LANGUAGE plpythonu OWNER TO 'karen'
```
The actual command needs to be 
```SQL
ALTER LANGUAGE plpythonu OWNER TO karen
```
I tried looking through the psycopg library to see if there was a way to change this but I didn't see one.

It actually bothers me to switch back and forth between these two different ways of making a SQL commands.  If I keep running into this issue I might just override the cursor cursor to hide this formatting change.

karenc: I think in that case, just add some comments so we know why it's done like that in the future. :)

cnxarchive/database.py
```diff
@@ -60,19 +64,33 @@ def _read_sql_file(name):
...
-    connection_string = settings[config.CONNECTION_STRING]
+    """Initialize the database from the given settings.
+       If a superuser is given in the settings then the
+       function will use the superuser to initialize the
+       database.
+    """
+    if settings['super-connection-string']:
+        connection_string = settings['super-connection-string']
+    else:
+        connection_string = settings['db-connection-string']
```
pumazi: Make sure you use the config constants for these (e.g. ``config.CONNECTION_STRING``).

cnxarchive/tests/testing.py
```diff
@@ -140,3 +141,92 @@ def setUp(self, cursor):
 # right timezone (America/Whitehorse is -07 in summer and -08 in winter)
 os.environ['PGTZ'] = 'America/Whitehorse'
 os.environ['TZ'] = 'America/Whitehorse'
+
+
+class CustomTest(unittest.TestCase):
```
pumazi: This isn't a good name.

cnxarchive/database.py
rich-hart: I was wondering about this.  Why are we using ``config.CONNECTION_STRING``?  Is there a situation where the dictionary keys for accessing our connection strings change?

cnxarchive/database.py
rich-hart: ok! done!

cnxarchive/database.py
rich-hart: Actually I guess it makes sense to keep them all in the same place.  I'll update the pull request.

reedstrm:
In general, we need the 'provide a superuser credential' part of this, but the schema creation part needs to be factored into steps that need to be done as the standard,non-superuser role (rhaptos or cnxarchive) and steps that require superuser privilege - i.e. creation of functions that use plpythonu, which then need to be OWNed by the standard user. I think the proper design is to make the standard db connection, use 'SELECT CURRENT_USER' to collect the role name, then connect as a super  user, SET ROLE to the standard user, except for plpythonu functions, which then you SET ROLE RESET, to create and then ALTER ... OWNER TO standard user.

PR #314, opened 2015-04-16
Adds a database table for reporting service status messages to the frontend

pumazi:
The schema change and the added constants are required database changes.

I made some assumptions about displaying the messages. Basically, the messages that have a priority >= 3 will be shown 4 hours before they become active. Others are shown 2 hours before they become active.


reedstrm:
Hmm, I was sort of thinking that we'd just show all non-expired messages, and let the frontend decide which to display. This works, too, but it does make the API non-deterministic - what time you query it affects what results you get, not just what data is available. (I guess mine does, too, in that it changes when you pass an expiry, but it seems less ... surprising, perhaps?

reedstrm:
This now returns all future messages, allowing the frontend to do the display logic. Merging.

PR #315, opened 2015-04-17
Add compression to memcached values if > 1MB

### LINKED ISSUE #312: Search results are too large for memcache
This is lengthening the time of our search to the point where webview is timing out before the results from archive can be returned.

pumazi:
Fixes #312

This also adds the 'memcache-max-value-length' settings key,
which will take a bytes value and apply that as the memcache
max value length rather than the default.

### DIFF
--- a/cnxarchive/cache.py
+++ b/cnxarchive/cache.py
@@ -65,7 +65,8 @@ def search(query, query_type, nocache=False):
             cache_length = int(settings['search-long-cache-expiration'])
 
         # store in memcache
-        mc.set(mc_search_key, search_results, time=cache_length)
+        mc.set(mc_search_key, search_results, time=cache_length,
+               min_compress_len=1024*1024)  # compress when > 1MB
 
     # return search results
     return search_results


karenc:
:+1:

karenc:
Should we add the setting to development.ini?

pumazi:
Ross didn't see the need for it, but I thought it'd be better to have for later performance tweaking. Actually, it probably needs to do an int cast for it to really work.

reedstrm:
@pumazi this missed the point of our conversation. While it's an interesting change, we need the max value length set to max, i.e. 128MB otherwise search results just aren't going to get cached. The missing API parameter is min_compress_len, which default to 0 i.e. no compression. That should probably be any positive value, say 1MB, but doesn't really need a config param.

reedstrm:
We're talking past each other. First off, memcached (the server) will complain and refuse to start with a setting higher than 128MB (-I switch):
```$ memcached -h
...
-I            Override the size of each slab page. Adjusts max item size
              (default: 1mb, min: 1k, max: 128m)
...
$ memcached -I 256m
Cannot set item size limit higher than 128 mb.
```

Second, we're already setting the stored size that high. Making it configurable is ok, but does not fix #312 - turning on compression will (help). Can we get a patch that does that, please?

PR #317, opened 2015-04-21
make fields match frontend

reedstrm:
frontend uses plural 'starts' and 'ends' . This has the advantage that 'end' is a postgresql keyword, do needs special quoting when used. ends is not.


pumazi:
You'll need to change the extras test.

pumazi:
I ran the tests locally. All looks good.

PR #318, opened 2015-04-28
Add the by-nc-sa license

pumazi:
See [Add NC-SA license option to Rewrite](https://trello.com/c/fHwODL3r)


cnxarchive/sql/get-module-metadata.sql
```diff
@@ -19,7 +19,7 @@ FROM (SELECT
   iso8601(m.created) AS created, iso8601(m.revised) AS revised,
   m.stateid, m.doctype,
   (SELECT row_to_json(license) AS license FROM (
-        SELECT l.code, l.version, l.name, l.url
+        SELECT l.code AS abbr, l.version, l.name, l.url
```
reedstrm: ISTM the term of art for CC license short codes is that: short code, or code, for short. Oh well.

pumazi:
@reedstrm Most definitely. It was a bad naming choice that happens to have come from authoring.

Perhaps we should change it to code in both authoring and archive?

webview doesn't use the value. I was trying to prevent changing authoring. This seemed like the easier thing to change, but it's also not the thing in a production environment.

PR #321, opened 2015-05-04
[TESTING] temp. revert of use of uuids instead of page numbers

reedstrm:
PR to get travis to run tests for me


pumazi:
Why are we doing this again? My brief recollection of why this is being done has to do with fixing something on production. If that's the case, shouldn't this PR be against the production branch rather than master?

edwoodward:
I think Michael is correct.

reedstrm:
Once these pass, I'll repoint to production.

reedstrm:
Closing in favor of PR against production, Like @pumazi  said.

PR #324, opened 2015-06-11
Eliminate link reference resolution

pumazi:


cnxarchive/transforms/resolvers.py
```diff
@@ -361,6 +361,9 @@ def fix_anchor_references(self):
         # Catch the invalid, unparsable, etc. references.
         bad_references = []
 
+        # XXX (10-Jun-2015) production only fix to resolve a trigger problem.
```
reedstrm: This skips the entire anchor transform - the /resources part still seems of use, to me. I was expecting to only skip if it's a MODULE_REFERENCE, perhaps even only if there's a collection_uuid involved.

pumazi:
A better solution can be found at #326.

PR #326, opened 2015-06-11
(For production) Fix reference resolution for page-in-book references 

pumazi:


pumazi:
The implementation of this for master can be found in #325.

reedstrm:
Very nice, merging.

PR #331, opened 2015-07-24
Added robots.txt script, removed blocked pages from Sitemap

cburk:
Added a script to generate a robots.txt file based on a list of blocked sites.

Created a function called equivalent to be used in Sitemap to ensure that none of the pages being added are equivalent to ones in the list of blocked sites mentioned above.


cnxarchive/views.py
```diff
@@ -36,6 +39,10 @@
 
 logger = logging.getLogger('cnxarchive')
 
+pages_to_block = ['legacy.cnx.org', '/lenses', '/*/pdf$', '/*/epub$', '/*/complete$',
```
pumazi: This is a constant. The convention for that is to make it all capital letters.

pumazi:
Why did you choose to make your own form of matching rather than use regular expressions?

cnxarchive/views.py
```diff
@@ -286,6 +293,67 @@ def _get_content_json(environ=None, ident_hash=None, reqtype=None):
...
+    weekdays = {0:'Mon', 1:'Tue', 2:'Wed', 3:'Thu', 4:'Fri', 5:'Sat', 6:'Sun'}
+    months = {'01': 'Jan', '02': 'Feb', '03': 'Mar', '04': 'Apr', '05': 'May', '06': 'Jun', '07': 'Jul', '08': 'Aug', '09': 'Sept', '10': 'Oct', '11': 'Nov', '12': 'Dec'}
+    datetime_str = datetime.__str__()
+
+    day = datetime_str.split()[0].split('-')[2]
+    month = datetime_str.split()[0].split('-')[1]
+    year = datetime_str.split()[0].split('-')[0]
+    time = datetime_str.split()[1][:8]
+
+    return weekdays[datetime.weekday()] + ', ' + day + ' ' + months[month] + ' ' + year + ' ' + time + ' GMT'
```
pumazi: There are better ways to format a string. Have you looked at https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior ?

pumazi:
The robots.py file needs to have 4 space indentation, rather than 2. This might seem like stylistic nitpicking, but it is a suggestion to prevent future editing from raising an ``IndentationError``.

cnxarchive/views.py
```diff
@@ -688,3 +712,27 @@ def sitemap(environ, start_response):
     headers = [('Content-type', 'text/xml')]
     start_response(status, headers)
     return [xml()]
+
+def robots(environ, start_response):
+    """
+    Returns a robots.txt file
+    """
+    robotsDotTxt = Robots()
```
karenc: Please use underscores instead of camel case variables in python.

pumazi:
This just needs a test or two.

cnxarchive/views.py
```diff
@@ -681,10 +709,46 @@ def sitemap(environ, start_response):
...
+        'Slurp': '10'
+        }
+
+    for bot in bot_delays:
+        robots_dot_txt.add_bot(bot, bot_delays[bot], PAGES_TO_BLOCK)
+
+    status = '200 OK'
+    gmt = timezone('GMT')
+    cur_time = datetime(1, 2, 3)
+    cur_time = cur_time.now(gmt) + timedelta(hours=-14)
```
karenc: why do we do this?

cnxarchive/views.py
```diff
@@ -681,10 +709,46 @@ def sitemap(environ, start_response):
...
+        'BecomeBot': '20',
+        'Slurp': '10'
+        }
+
+    for bot in bot_delays:
+        robots_dot_txt.add_bot(bot, bot_delays[bot], PAGES_TO_BLOCK)
+
+    status = '200 OK'
+    gmt = timezone('GMT')
+    cur_time = datetime(1, 2, 3)
```
karenc: This line is not necessary, you should be able to do `datetime.now(gmt)` on the next line.

cnxarchive/views.py
karenc: the - 14 hours thing

karenc:
Ready for review again.

pumazi:
@reedstrm Can you verify the contents of the robots.txt? I'm assuming that you probably came up with the required info in there. Please merge if you are satisfied with it.

reedstrm:
The main issue here is that this disallows a bunch of URL patterns that are specific for legacy. We've tweaked the existing robots.txt so that legacy is Disallow: *, in any case. Having robots.txt part of the python code makes sense only in that developers who add routes would then need to decide if those routes (or parts of them) should be excluded from search engine spidering. The existing list are things that are expensive on legacy and accumulated over the years as things to keep the spiders out of. We've now blocking all of that, already. This list needs to be changed to reflect URLs that would appear in rewrite (actually, in the prerender output) that we want to keep spiders out of.  I haven't spent any brain cycles thinking about what URLs we may want to block. Anything having to do with authoring. Login/logout. Ideas?

More substantial criticism: while the list of routes/URLs to block seems appropriate to put in code (since you can only add routes by changing code), the list of spiders/bots to act on does not. That should be a config setting. Not sure the delay needs to be alterable per bot: that there are two diff. values is probably accident of history.

karenc:
Is @cburk going to look at this?

edwoodward:
Yes, he is out today.

reedstrm:
I need to amend my comment: I was mistaken in thinking that all legacy links had been redirected to legacy.cnx.org. We allow a lot of them through to allow for legacy editing. While the URL patterns need reconsidering, they're not useless as it stands.

cburk:
After discussion in scrum, we decided to reopen this pull request and change the blocking of a couple pages.  If there are no major issues with this, it should be good to merge.

reedstrm:
looks good, merging

reedstrm:
Ah, this is against production (which is good) but we need a version against master, as well. @cburk do we still have some of your time this week?

PR #334, opened 2015-07-30
Migration to the Pyramid web framework.

karenc:
Clean up commits to follow in another PR.


reedstrm:
is this in flight, or ready for review?

karenc:
It is ready for review

pumazi:
I think we need to think about how this will affect deployment before we merge it. I only say this because production is far enough from master as it is. Adding this on top of it could be too much.

reedstrm:
Yay for devops thinking @pumazi. Agreed.

karenc:
Let's think about how to get production to current master.

reedstrm:
This has passed checks, I think we should deploy it to dev and smoke out any issues.

pumazi:
We may need to address the change in #355. It's something to look into anyway.

cnxarchive/__init__.py
```diff
@@ -33,114 +28,45 @@
     'content-type',
     ]
 
+# XXX (25-Jul-2015) Should be removed ASAP
 def get_settings():
```
pumazi: Let's add an issue for this. It's one of those things that shouldn't be left hanging around.

pumazi:
:+1: We should add issues to remove the settings getter/setter functions and the application helpers. I can probably take care of these later this week or over the weekend.

I think we are ready to part the Red Sea. 

@reedstrm Feel free to merge it whenever you're ready. Thanks for letting me look over it first. :checkered_flag:

PR #339, opened 2015-08-13
robots.txt from production for master

reedstrm:
Added robots.txt script, removed blocked pages from Sitemap. Pretty sure this brings over all the robots.txt code that was merged to production


karenc:
We're going from 1.2.0 to 1.6.0 because we tagged without updating the version in setup.py ...

PR #340, opened 2015-08-14
[Production] logging and license changes

reedstrm:


pumazi:
:+1: Looks good. Though, I'm not sure why the hostname change in the tests was necessary.

reedstrm:
I didn't like the hardcoded cnx.org in the Robots init, decided to make it explicit. (e.g. the test system was generating cnx.org instead of demo.cnx.org) Seemed to be relying on the potentially dodgy default of 127.0.0.1. And since I was passing it explicitly from HOST_NAME for one case, I made it consistent.

PR #341, opened 2015-08-25
Update cnx-archive to use python 3 (but still compatible with python 2)

karenc:
This is based on #352.


karenc:
Rebased.

rich-hart:
@reedstrm @pumazi @mwharrison @karenc  I think that we need to get this branch merged into master ASAP.  It's an important update that is only going to get hard the longer we wait.  Right now, the most resent commits to master are not python3 compatible, so this branch is going to need another update.   I've tested the current state of this branch on my local machine with python 2 and 3 both work great.  What do we need to do to get this done?

karenc:
@rich-hart This depends on the inlining triggers PR #352.  There are changes from the short id PR #366 that we need to fix before we can continue with this.

rich-hart:
Ok, I'm going to rebase this branch against #352 first.

rich-hart:
This is based on #367

karenc:
Going to submit a new PR for this.

Issue #335, opened 2015-07-31
Author names are not in the same order as on Legacy


amberwebb:
This was reported by user in Trac ticket 13213

The order of the authors is different between legacy and rewrite. it could be a display issue or a data issue.

legacy: https://legacy.cnx.org/content/col11569/latest/

Rewrite: http://cnx.org/contents/db89c8f8-a27c-4685-ad2a-19d11a2a7e2e@21.1:1/Principles_of_Biology

amberwebb:
Original issue: https://github.com/Connexions/webview/issues/1096

edwoodward:
This is fixed

PR #342, opened 2015-08-27
Authors out of order based on list stored in the database, fixes issue #335

mwharrison:
The authors are out of order based on the list in the database.
Implemented index based array sorting in the sql. Updated existing
tests to match new return order (coverage is acceptable).


reedstrm:
For devops purposes, it's important to note that this is a schema (adding the function def) Ideally, include the commands necessary to bring a DB up-to-date is nice. In this case:

rich-hart:
I see that some of the ``CONTENT_METADATA`` fields were changed in the ``test_view.py`` unit tests, but its not clear to me which specific test is being used to insure the ordering of authors.   Would mind mentioning which test demonstrates the behavior your looking for?

mwharrison:
test_collection_content will fail if the authors are returned out of order

rich-hart:
Looks good then.  Merging...

PR #345, opened 2015-08-31
Explode the schema file into smaller bitesize pieces

pumazi:


pumazi:
I'm looking for opinions on this structure. Does this help readability and organization???

Also, though I haven't implemented it yet, it is possible to compile all of these into a single file with something like:

```python
from cnxarchive.database import get_schema
with open('schema.sql', 'wb') as file:
    file.write(get_schema())
```

notify: @rich-hart @karenc @reedstrm

reedstrm:
As you may (or may not) recalI,  started the process of splitting functions out w/ the tree* functions and structure. I like the separate mainfest file. There are some extra bits (I assume the modifier func. is a just in case future proofing?) Not sure how much this helps with incremental deployment, however. It'll help in the cases of new, added functionality that is well contained in its own files (e.g. the publishing tables). It won't help much with adding a field, or a trigger, for example. But I guess we can't solve all problems at once. :-)
A single file is really needed, as long as we keep the CLI db initializer: it's easy enough to initialize a throw-away DB and pg_dump it. Actually, I think that's going to be the primary incremental update method: init, then dbtoyaml, then yamltodb against the to-be-updated db.
So, :+1:

cnxarchive/database.py
```diff
@@ -62,20 +48,62 @@ def _read_sql_file(name):
...
+    return joiner.join(items)
+
+
+def get_schema():
+    manifest_filepath = os.path.join(DB_SCHEMA_DIRECTORY,
+                                     SCHEMA_MANIFEST_FILENAME)
+    schema_manifest = _read_schema_manifest(manifest_filepath)
+
+    # Wrap the files in comments that say it's origin file.
+    file_wrapper = lambda f, c: u"-- FILE: {0}\n{1}\n-- \n".format(f, c)
```
reedstrm: could use the description field from the manifest here, as part of the separator.

reedstrm: Actually, not sure why we'd want to create one giant string: starting a transaction, and executing each file's contents separately would allow us to bail more cleanly on any issues.

pumazi:
@reedstrm can we merge this as is and progressively update as we go? (I'm only suggesting this because I'm not going to be working on it for awhile, since I'm changing focus to biglearn.)

pumazi:
@karenc is suggesting I do a schema comparison to be sure this doesn't hose everything. I'll get that done shortly.

karenc:
I've compared the schema between master and boom using Pyrseas [dbtoyaml](http://pyrseas.readthedocs.org/en/latest/dbtoyaml.html) and the schema is exactly the same.

:+1:

PR #346, opened 2015-09-01
Name searching match

mwharrison:
Increased author search test case to include the specific issue, unable to match on partial author searches, or authors with middle names. Added space replacements in the query.

### DIFF
--- a/cnxarchive/sql/search/author.part.sql
+++ b/cnxarchive/sql/search/author.part.sql
@@ -9,7 +9,7 @@ SELECT
   %({0})s::text||'-::-author' as key
 FROM
   latest_modules AS m,
-  users AS u 
+  users AS u
 WHERE
   u.username = ANY (m.authors)
   AND
@@ -17,7 +17,7 @@ WHERE
    OR
    u.last_name ~* req(%({0})s::text)
    OR
-   u.full_name ~* req(%({0})s::text)
+   u.full_name ~* regexp_replace(req(%({0})s::text), ' ', '.+', 'g')
    OR
    u.username ~* req(%({0})s::text)
    )
--- a/cnxarchive/tests/test_search.py
+++ b/cnxarchive/tests/test_search.py
@@ -291,7 +291,7 @@ class SearchTestCase(unittest.TestCase):
             'username': 'jmiller',
             'first_name': 'Jill',
             'last_name': 'Miller',
-            'full_name': 'Jill M.',
+            'full_name': 'Jill S. Miller',
             'title': None,
             }
 
@@ -310,7 +310,9 @@ INSERT INTO users
     def test_author_search(self, cursor):
         # Test the results of an author search.
         user_info = self._add_dummy_user()
-        query_params = [('author', user_info['first_name'])]
+        # we want to make sure an author can be searched by first and last name
+        # even if they have a middle initial
+        query_params = [('author', 'Jill Miller')]
 
 
         # Update two modules in include this user as an author.


cnxarchive/sql/search/author.part.sql
```diff
@@ -17,7 +17,7 @@ WHERE
    OR
    u.last_name ~* req(%({0})s::text)
    OR
-   u.full_name ~* req(%({0})s::text)
+   u.full_name ~* regexp_replace(req(%({0})s::text), ' ', '.*.')
```
karenc: Is `.*` enough to do wildcard match?  Should we do a global replace?

karenc: Come to think about it, should we use `.+` instead?  To make sure when people search for `jill miller` we won't return `jillmiller`? (just in case)

mwharrison: Yeah, I guess that makes sense. Come to think about it, I can see where a global replace would be better - if someone has a duel first or last name (also, just in case). I'll update and issue another PR.

cnxarchive/sql/search/author.part.sql
```diff
@@ -9,15 +9,15 @@ SELECT
...
   u.username = ANY (m.authors)
   AND
-  (u.first_name ~* req(%({0})s::text)
+  (u.first_name ~* regexp_replace(req(%({0})s::text), ' ', '.+.')
    OR
-   u.last_name ~* req(%({0})s::text)
+   u.last_name ~* regexp_replace(req(%({0})s::text), ' ', '.+.')
    OR
-   u.full_name ~* req(%({0})s::text)
+   u.full_name ~* regexp_replace(req(%({0})s::text), ' ', '.+.')
```
karenc: I think we should either have `.+` (my preference) or `.*.`, not `.+.`, otherwise it wouldn't match "jill miller".  And why remove the global flag?

mwharrison: Sorry, just added that back - and changed to .+

cnxarchive/sql/search/author.part.sql
```diff
@@ -9,15 +9,15 @@ SELECT
   %({0})s::text||'-::-author' as key
 FROM
   latest_modules AS m,
-  users AS u 
+  users AS u
 WHERE
   u.username = ANY (m.authors)
   AND
-  (u.first_name ~* req(%({0})s::text)
+  (u.first_name ~* regexp_replace(req(%({0})s::text), ' ', '.+', 'g')
```
reedstrm: What's the purpose of replacing the spaces?

mwharrison: So that we can match a query on any part of the users full name.

reedstrm: I tested, your way is faster and more correct. Though I stand by only needing it on fullname.

karenc:
Let's squash all the commits before merging this :)

mwharrison:
Okay - that was my first time squashing commit, I feel like I did something wrong. I also noticed tests were failing because of some missing parentheses, so I fixed that. Did it work?

karenc:
That didn't work.  There are 8 commits in this PR.

An easy way to fix this is on your branch:

1. `git rebase master`
2. `git reset master`    (this removes all the commits in this branch)
3. `git commit -a`    (commit your changes, you might also want to fix those parentheses before you commit)
4. `git push -f origin HEAD`    (`-f` is necessary because you've removed commits)

PR #350, opened 2015-09-11
return 404 "not found" for the final-slash-with-no-id case as well

reedstrm:
JP noticed that http://archive.cnx.org/content and http://archive.cnx.org/content/ behaved differently. Fix the later to be like the former.

### DIFF
--- a/cnxarchive/utils.py
+++ b/cnxarchive/utils.py
@@ -41,10 +41,11 @@ def split_ident_hash(ident_hash, split_version=False):
     if HASH_CHAR not in ident_hash:
         ident_hash = '{}@'.format(ident_hash)
     split_value = ident_hash.split(HASH_CHAR)
-    if split_value[0] == '':
-        raise ValueError("Missing values")
 
     try:
+        if split_value[0] == '':
+            raise ValueError("Missing values")
+
         id, version = split_value
     except ValueError:
         raise IdentHashSyntaxError(ident_hash)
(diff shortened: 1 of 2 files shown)


karenc:
Is there a test we can add?

reedstrm:
Worse than hat, there _is_ a test, and I failed to correct it. Will fix now.

reedstrm:
Empty value is specc'ed to return 'ValueError' all other bad ID cases return 'IdentHashSyntaxError'. Either this util function should return 'IdentHashSyntaxError', or all calls to it that are guarded w/ try/except need to catch either error. I  prefer the former, on symplicity grounds: if some caller needs to notice an empty ID, they can test for it there.

cnxarchive/utils.py
```diff
@@ -41,10 +41,11 @@ def split_ident_hash(ident_hash, split_version=False):
...
         ident_hash = '{}@'.format(ident_hash)
     split_value = ident_hash.split(HASH_CHAR)
-    if split_value[0] == '':
-        raise ValueError("Missing values")
 
     try:
+        if split_value[0] == '':
+            raise ValueError("Missing values")
+
         id, version = split_value
```
pumazi: No, this isn't right. This is a ValueError. The value is missing so, it's legitimately an error with the value, not the syntax.

pumazi: What I'm saying is that you shouldn't catch it and raise it as an identhashsyntaxerror.

karenc:
I was hoping that you could add a test for `content` vs `content/`..?

reedstrm:
merged via force push

PR #352, opened 2015-09-13
Inline most trigger code, remove plpydbapi and move transforms code

karenc:
This pull request will allow us to:
 - not have to install cnx-archive (and other packages not used in triggers) globally,
 - remove dependency on plpydbapi (and remove our [fork](https://github.com/Connexions/plpydbapi)),
 - be more sure about the code in the trigger (by inlining the code for all triggers except add module files trigger).

TODO: I've added some sed commands in .travis.yml to add cnx-cnxml-transforms, lxml and rhaptos.cnxmlutils to sys.path in the triggers.  If we could add this when we read the schema in or something, that'll be great.

=====

- Inline code for assign version default trigger
  
- Inline code for assign uuid default trigger
  
- Inline code for upsert document acl trigger
  
- Inline code for upsert users from legacy publication trigger
  
- Inline code for insert users for optional roles trigger
  
- Inline code for republish module trigger
  
- Inline code for assign moduleid default trigger
  
- Move cnxarchive.transforms to cnx-cnxml-transforms
  
- Inline some code for add module file trigger
  
- Reimplement add module files with plpy, remove plpydbapi dependency
  
- Stop using system python on travis
  - cnx-archive is no longer imported from postgres python triggers
  - cnx-cnxml-transforms is imported from postgres python triggers but can
    be made available by appending to sys.path within the triggers


pumazi:
This work was done in an effort to better manage triggers. The work seen here is an alternative to things we've discussed around making our triggers "virtualenv aware."

We came up with a game plan for [addressing 2 of the three problems/issues](https://gist.github.com/pumazi/b204740439e684fbe982#problems). Unfortunately, we cannot resolve all the problems in one swath. We started down a path of looking at solutions that would [remove imports from packages altogether](https://gist.github.com/pumazi/b204740439e684fbe982#dont-import-packages-from-the-triggers) and [moving the triggers into a separate package](https://gist.github.com/pumazi/b204740439e684fbe982#separate-the-triggers-into-their-on-package). Some of the triggers fit into one solution while not into the other. We settled on middle ground (this work) to [separate the trigger logic into it's own package](https://gist.github.com/pumazi/b204740439e684fbe982#separate-the-trigger-logic-into-its-own-package), which solves some of the issues. The remaining issue will likely be addressed when legacy is finally retired.

pumazi:
:+1:  :100:

pumazi:
@reedstrm Any thoughts or opinions on this work?

pumazi:
We'll need to adjust one of the triggers to include the change from #363.

karenc:
Rebased and included changes from #363

pumazi:
What are the chances this will get merged? :8ball:

I'm getting the feeling this is unwanted, but nothing is being said... :ear: @reedstrm @rich-hart @mwharrison @edwoodward

reedstrm:
Let me put the same comment here. I like this work, and I think it keeps the best part of the import-core-of-trigger (being able to update code w/o exclusive locks on the DB), but takes out some of the annoyances (needing the whole archive server package on machines that only run the db). Primary reason this has been held up is the short timeline and hard deadline for the Concept Coach work, in particular the short_id work. Hang in there, we'll get this merged.

karenc:
Do we want to change the way this is implemented?  May be we no longer want to inline the code for triggers.  We could create another package for the other trigger code so it's not in archive?  (Transformation triggers are already in its own package).

rich-hart:
No, I think this is fine.  I do think that it would be nice to move the trigger code to is own directory.  But I don't think that it's necessary.   Plus I would like to keep the triggers associated with archive for now.

karenc:
Before we do more with this PR, please go back and fix all the comments raised in #366

karenc:
Going to submit a new PR for this.

PR #353, opened 2015-09-16
Production rebase

reedstrm:
creating PR to fire travis tests, and consider diff


reedstrm:
happened via a push -f. Not sure why github thinks there's still a diff.

PR #355, opened 2015-09-17
Yes we have no bananas for jp

reedstrm:


cnxarchive/utils/ident_hash.py
```diff
@@ -40,10 +40,10 @@ def split_ident_hash(ident_hash, split_version=False):
     if HASH_CHAR not in ident_hash:
         ident_hash = '{}@'.format(ident_hash)
     split_value = ident_hash.split(HASH_CHAR)
-    if split_value[0] == '':
-        raise ValueError("Missing values")
 
     try:
+        if split_value[0] == '':
+            raise ValueError("Missing values")
```
pumazi: I'm sure this works, but I strongly feel that it is incorrect. you shouldn't catch it and raise it as an ``IdentHashSyntaxError``, because the value is missing, which means there is no syntax to error on. Therefore, it's legitimately an error with the value (aka ValueError), not the syntax.

reedstrm:
we could argue the semantics of "is an empty string a syntax error" or not. Practically, it means having to catch two different errors at each calling place upstream, or guard all calls so as not to pass '' in.

reedstrm:
Having said that, the specific code seemed hacky  - raising and catching the same error.

cnxarchive/utils/ident_hash.py
karenc: Should we change the views to also handle `ValueError` and raise `HTTPNotFound`?

reedstrm:
Changes this to a change in routing only, so /contents and /contents/ are treated the same: not matched.

reedstrm:
The list of commits in the first message is no longer valid. Redid as a routing change, w/ tests

pumazi:
Can we change the PR title? Just in case we need to search for this later... :sweat_smile:

pumazi:
:+1: Ok! :clap:

I just touched up the styling.

reedstrm:
suggest a name? "404 contents route final slash"?

PR #356, opened 2015-09-17
Parentauthors order [production]

reedstrm:


reedstrm:
Aaarrg! I've fixed that test data set sooo many times. Digging in, the latest_abstract data in the test data is wrong: it should have abstractIds that match those in modules.

PR #360, opened 2015-09-28
Use pyramid route_path instead of hardcoded paths

karenc:
- Implement a route pregenerator that allows incomplete route args

  The default route generator throws a KeyError if any variable in the
  path pattern is not defined.

  The pregenerator makes sure we can generate a path using
  request.route_path even if we don't have all the variables.

  For example, instead of having to do this:
      `request.route_path('resource', hash=hash, ignore='')`
  it's possible to do this:
      `request.route_path('resource', hash=hash)`

- Change all hardcoded paths to use request.route_path


pumazi:
:+1: d

PR #361, opened 2015-09-28
Production pyramid merge

reedstrm:
merging the pyramid port from master to production


pumazi:
We need to **rename the production branch to spike-1.6.x** and/or scrap the production branch altogether. If you insist on keeping a production branch, I would suggest you pave over (force push master onto) the existing production branch.

reedstrm:
@pumazi would love too, but we still need the page# based ref, since we haven't gotten that fix approved for completion of development.

reedstrm:
THe failed test above has this signature:

reedstrm:
I'm treating this PR as documenting my rollout, so I'm going to merge it myself.

PR #364, opened 2015-10-13
WIP: In book search

mwharrison:


karenc:
You should always use `git rebase master` instead of `git merge master`.  (There should never be a merge master into your-development-branch commit)

mwharrison:
Is there anyways to undo that?

karenc:
You can just do `git rebase master` now and it will get rid of the merge commit.

reedstrm:
right this is a WIP branch so I can comment directly on code. Wish github allowed that just on a branch. :-)

reedstrm:
But cleaning it up "the right way" is a good lesson for us all, as well.

cnxarchive/__init__.py
```diff
@@ -53,6 +53,7 @@ def add_route(name, path, *args, **kwargs):
     add_route('export', '/exports/{ident_hash}.{type}{ignore:(/.*)?}')  # cnxarchive.views:get_export
     add_route('content-extras', '/extras/{ident_hash}')  # cnxarchive.views:get_extra
     add_route('search', '/search')  # cnxarchive.views:search
+    add_route('in-book-search', '/search/book/{ident_hash}')  # cnxarchive.views:in-book-search
```
reedstrm: This should 'just work' with /search/{ident_hash}' , since the other search route never has a path component at all, only params

cnxarchive/sql/get-in-book-search.sql
```diff
@@ -0,0 +1,40 @@
...
+
+-- arguments: uuid:string, major_version:int, minor_version:int, search_term:string
+WITH RECURSIVE t(node, title, path,value, depth, corder) AS (
+SELECT nodeid, title, ARRAY[nodeid], documentid, 1, ARRAY[childorder]
+FROM 
+  trees tr, 
+  modules m
+WHERE 
+  m.uuid::text = %(uuid)s AND
+  m.major_version = %(major_version)i AND  m.minor_version = %(minor_version)i AND
```
reedstrm: The only supported format specifier is 's', so this should be %(major_version)s Psycopg does the right thing if you pass it an integer:

cnxarchive/views.py
```diff
@@ -558,6 +631,15 @@ def search(request):
         page = None
     if page is None or page <= 0:
         page = 1
+    try:
```
reedstrm: let's not do this here at all for now.

karenc:
I comment on it because there's never any reason to use `git merge` instead of `git rebase

cnxarchive/views.py
```diff
@@ -516,6 +516,79 @@ def get_export(request):
...
+            u'per_page': DEFAULT_PER_PAGE,
+            u'page': 1,
+            },
+        u'results': {
+            u'items': [],
+            u'total': 0,
+            u'limits': [],
+            },
+    })
+
```
reedstrm: This needs something like:
    args = request.matchdict
    ident_hash = args['ident_hash']
    id, version = split_ident_hash(ident_hash)
    if version:
        if '.' in version:
            major_version, minor_version = [int(v) for v in version.split('.')]
        else:
            major_version = int(version)
            minor_version = None

Then, down inside the cursor scope/context:
            if not version:
                redirect_to_latest(cursor, id, 'in-book-search', route_args=request.params.copy())

mwharrison:
Example result: https://gist.github.com/mwharrison/8b56c7c7b6299db62a4b

PR #366, opened 2015-10-26
shortid changes

rich-hart:
Credit: @reedstrm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,10 +42,9 @@ before_script:
   - sudo -u postgres psql -d postgres -c "CREATE USER cnxarchive WITH SUPERUSER PASSWORD 'cnxarchive';"
   # Set up the database
   - sudo -u postgres createdb -O cnxarchive cnxarchive-testing
-  # Test pep8 - eat failure code for now
   - pip install pep8
-  - (pep8 --exclude=tests *.py cnxarchive/; /bin/true)
-  - (pep8 --max-line-length=200 cnxarchive/tests; /bin/true)
+  - pep8 --exclude=tests *.py cnxarchive/
+  - pep8 --max-line-length=200 cnxarchive/tests
 
 script:
   # This is the same as `python -m unittest discover` with a coverage wrapper.
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,10 @@ before_script:
   - sudo -u postgres psql -d postgres -c "CREATE USER cnxarchive WITH SUPERUSER PASSWORD 'cnxarchive';"
   # Set up the database
   - sudo -u postgres createdb -O cnxarchive cnxarchive-testing
+  # Test pep8 - eat failure code for now
   - pip install pep8
-  - pep8 --exclude=tests *.py cnxarchive/
-  - pep8 --max-line-length=200 cnxarchive/tests
+  - (pep8 --exclude=tests *.py cnxarchive/; /bin/true)
+  - (pep8 --max-line-length=200 cnxarchive/tests; /bin/true)
 
 script:
   # This is the same as `python -m unittest discover` with a coverage wrapper.
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -42,6 +42,7 @@ SQL = {
     'get-resourceid-by-filename': _read_sql_file('get-resourceid-by-filename'),
     'get-tree-by-uuid-n-version': _read_sql_file('get-tree-by-uuid-n-version'),
     'get-module-versions': _read_sql_file('get-module-versions'),
+    'get-module-uuid': _read_sql_file('get-module-uuid'),
     'get-subject-list': _read_sql_file('get-subject-list'),
     'get-featured-links': _read_sql_file('get-featured-links'),
     'get-users-by-ids': _read_sql_file('get-users-by-ids'),
@@ -111,7 +112,7 @@ def initdb(settings):
 
 def get_module_ident_from_ident_hash(ident_hash, cursor):
     """Returns the moduleid for a given ``ident_hash``."""
-    uuid, (mj_ver, mn_ver) = split_ident_hash(ident_hash, split_version=True)
+    uuid, (mj_ver, mn_ver), id_type = split_ident_hash(ident_hash, split_version=True)
     args = [uuid]
     stmt = "SELECT module_ident FROM {} WHERE uuid = %s"
     table_name = 'modules'
@@ -137,7 +138,7 @@ def get_tree(ident_hash, cursor):
     """Given an ``ident_hash``, return a JSON representation
     of the binder tree.
     """
-    uuid, version = split_ident_hash(ident_hash)
+    uuid, version, id_type = split_ident_hash(ident_hash)
     cursor.execute(SQL['get-tree-by-uuid-n-version'],
                    (uuid, version,))
     try:
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -112,7 +112,8 @@ def initdb(settings):
 
 def get_module_ident_from_ident_hash(ident_hash, cursor):
     """Returns the moduleid for a given ``ident_hash``."""
-    uuid, (mj_ver, mn_ver), id_type = split_ident_hash(ident_hash, split_version=True)
+    uuid, (mj_ver, mn_ver), id_type = split_ident_hash(
+        ident_hash, split_version=True)
     args = [uuid]
     stmt = "SELECT module_ident FROM {} WHERE uuid = %s"
     table_name = 'modules'
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -112,7 +112,7 @@ def initdb(settings):
 
 def get_module_ident_from_ident_hash(ident_hash, cursor):
     """Returns the moduleid for a given ``ident_hash``."""
-    uuid, (mj_ver, mn_ver), id_type = split_ident_hash(
+    uuid, (mj_ver, mn_ver) = split_ident_hash(
         ident_hash, split_version=True)
     args = [uuid]
     stmt = "SELECT module_ident FROM {} WHERE uuid = %s"
@@ -139,7 +139,7 @@ def get_tree(ident_hash, cursor):
     """Given an ``ident_hash``, return a JSON representation
     of the binder tree.
     """
-    uuid, version, id_type = split_ident_hash(ident_hash)
+    uuid, version = split_ident_hash(ident_hash)
     cursor.execute(SQL['get-tree-by-uuid-n-version'],
                    (uuid, version,))
     try:
--- a/cnxarchive/scripts/hits_counter.py
+++ b/cnxarchive/scripts/hits_counter.py
@@ -93,7 +93,7 @@ def main(argv=None):
     with psycopg2.connect(connection_string) as db_connection:
         with db_connection.cursor() as cursor:
             for ident_hash, hit_count in hits.items():
-                id, version, id_type = split_ident_hash(ident_hash)
+                id, version, id_type = split_ident_hash(ident_hash,return_type=True)
                 if id_type == CNXHash.FULLUUID:
                     cursor.execute(SQL_GET_MODULE_IDENT_BY_UUID_N_VERSION,
                                    (id, version))
--- a/cnxarchive/scripts/hits_counter.py
+++ b/cnxarchive/scripts/hits_counter.py
@@ -17,7 +17,7 @@ import gzip
 
 import psycopg2
 from .. import config
-from ..utils import split_ident_hash
+from ..utils import split_ident_hash, CNXHash
 from ._utils import create_parser, get_app_settings_from_arguments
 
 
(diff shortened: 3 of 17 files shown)


.travis.yml
```diff
@@ -42,9 +42,10 @@ before_script:
   - sudo -u postgres psql -d postgres -c "CREATE USER cnxarchive WITH SUPERUSER PASSWORD 'cnxarchive';"
   # Set up the database
   - sudo -u postgres createdb -O cnxarchive cnxarchive-testing
+  # Test pep8 - eat failure code for now
   - pip install pep8
-  - pep8 --exclude=tests *.py cnxarchive/
-  - pep8 --max-line-length=200 cnxarchive/tests
+  - (pep8 --exclude=tests *.py cnxarchive/; /bin/true)
+  - (pep8 --max-line-length=200 cnxarchive/tests; /bin/true)
```
pumazi: Part of the point of these lines was to fail when pep8 problems are found. :-1:

rich-hart: oops, forgot to leave pep8 validation back in.   I'll fix it

cnxarchive/database.py
```diff
@@ -111,7 +112,7 @@ def initdb(settings):
 
 def get_module_ident_from_ident_hash(ident_hash, cursor):
     """Returns the moduleid for a given ``ident_hash``."""
-    uuid, (mj_ver, mn_ver) = split_ident_hash(ident_hash, split_version=True)
+    uuid, (mj_ver, mn_ver), id_type = split_ident_hash(ident_hash, split_version=True)
```
pumazi: Why do we need the id_type returned from every occurrence of the split?

I have a feeling this is going to **rot into a pungent code smell**. Perhaps we should treat all ident-hashs as invalid if they aren't the short version?

We will only handle the uuid based ids in known places, right? One of those being during route matching for redirecting uuid based URIs to short-id based URIs.

pumazi:
Why do we need the ``id_type`` returned from every occurrence of the split?

I have a feeling this is going **to rot into a pungent code smell**. Perhaps we should treat all ident-hashs as invalid if they aren't the short version?

We will only handle the uuid based ids in known places, right? One of those being during route matching for redirecting uuid based URIs to short-id based URIs. I think that might actually be the only place, no?

pumazi:
I'm not sure what the use of the CNXHash equality methods are about. Can you speak to why these are needed?

Wouldn't it be better to use Pythonic comparison by implementing the comparison methods (e.g. [``__eq__``](https://docs.python.org/2/reference/datamodel.html#object.__eq__))?

rich-hart:
@pumazi 

*Why do we need the id_type returned from every occurrence of the split_ident_hash? We will only handle the uuid based ids in known places, right?*

I think that this was a preference of @reedstrm to have the id_type returned by split_ident_hash.  I'll have to double check with him.

*Wouldn't it be better to use Pythonic comparison by implementing the comparison methods (e.g. __eq__)?*

I originally wanted to implement it that way.  I just didn't get a chance to overwrite the __eq__ method.  The reason why I went through the trouble of the equality and similarity methods so that there would be a generalize way of comparing different type of hashes like, UUID, base64 encoding of a uuid, short base64 encoding of a uuid, etc.  The equality function tests whether two id's are the same, but just in a different encoding.  The similarity function tests whether a shortid hash **might** be equivalent to a base64hash, or uuid, but it's not certain since the shortid doesn't have the same amount of information as the others.

Functions like 'similarity' might not be used right now but I wanted to include it in this pull request just so it'll be a part of our utilities library.

karenc:
I'm not sure if we need to know whether a uuid and short id are equivalent or similar.  The way I think is once we generate a short id, we can forget how we got it (derived from the uuid).  We can just look up the short id from the database.  I don't think we even need to validate it.  May be I don't understand all the requirements for the short id?

rich-hart:
@karenc I put the equality and similarity functions in because our short ids have less information associated associated them than our uuids, and I wanted to account for that difference.  We probably could get by without needing to ever call the similarity function, but I wanted to add it to my solution because i felt it was an important mathematical detail.

Also I added that validate function because I wanted to be able to say, whether or not a short_id we received was consistent with our hashing conventions.   That way, if we received a hash string that was only 1 character long, we would know that it wasn't a valid short id before making a query that would return our entire database.

karenc:
We can add a length requirement if the id is not a uuid.  I just think how we generate the short id shouldn't matter as long as they are unique in the database.  There's no need to enforce some kind of relationship between the short id and the uuid?

karenc:
I missed the fact that we're not storing the short id in the database.  Why aren't we storing it in the database?

rich-hart:
Can you elaborate on what you mean by 'enforced relationship'?

rich-hart:
*Why aren't we storing it in the database?*
Because @reedstrm explained that we were still going to hold onto our uuid standard for the forceable future.  So if we are going to store all our UUIDs in the database it's easier, about just as fast, to calculate short hashes as we need them.

pumazi:
Rather than redirecting per-view using this ``id_type``, we can generalize the route mapper to route uuid based ident-hashes to a view that rewrites the uuid based ident-hash to a short-id ident-hash for all occurrences. That in psuedo code would look something like:

```python
def redirect_uuid_ident(request):
    matchdict = request.matchdict.copy()
    ident_hash = matchdict['ident_hash']
    matchdict['ident_hash'] = change_to_short_id(ident_hash)
    route_name = <somewhere in the request object>
    global uuid_routes_mapper
    return httpexceptions.HTTPFound(request.route_url(uuid_routes_mapper[route_name], **matchdict)
```

Or something like that...

reedstrm:
Short id's are directly derived from the 128bit number underneath the uuid. We don't store it separately because it's easy enough to generate, and doesn't lock us into this one implementation. They will be unique, due to the index on latest_modules(short_id(uuid)).

Chose to stick with full uuids as the standard interface for archive retrieval of content. Short Id's are only used to generate redirects to the equivalent full id.  Put the id type where we did since that's the one method that needs to know about it, since it 1: validates the uuid, and 2: sets up the 'lack of version, need to redirect' path. Which now has two cases for redirect: short id, and lack of version.

Let me repeat that: short Ids are for human consumption only: all content fetching/creation etc. will happen via uuid@version hashes. The only reason we need it in the REST API at all is sometimes webview will be handed a URL w/ only short ids in it, so has to ask  the DB/archive what the full id is.

We _might_ find we need similar code in authoring, so moving the CNXHash to  a shared code base might be a good idea. In some sense, it belongs with cnx-epub, as part of the models. As it stands, there's probably some YAGNI code in there (I also am not sure about the similarity stuff) - Rich looked at this bottom-up, and designed a full-featured type, not the bare minimum to support the redirects.

Aside: rotten code smell? in bold? Really?

karenc:
I thought it would be nice to just use the short ids and only use uuids for redirecting old urls.

Can I just ask if we're doing anything in case there's a short id collision?  Or is that just not going to happen?

pumazi:
Having a better understanding of the direction these things are being used is useful. (I'm guessing that needs documented somewhere?) I wasn't aware that the short-id was only to be used for human readability.

Why not just use the short-id everywhere? Is there any reason we should be using both id forms externally?

My comments on the ``id_type`` stands. I think it's better to assume you have the correct thing instead of checking each time to make sure it is correct. Various pieces of code will needlessly acquire if statements instead of handling the problem at the routing layer and therefore in a single location.

reedstrm:
BTW, that's a weird test to fail. I tried it locally, and it passes. Retiggering travis.

reedstrm:
re: documentation - agreed, and apologies. As part of pep8ification, I turned on flake8 in my editor, and it griped a lot about missing documentation comments, everywhere. As I write them, I realized that this is exactly the sort of info @pumazi and @karenc need.  I plan to work through them all today and add that to this PR.

Re: short vs. long ids, primarily, we need to keep full uuids for external APIs for the reason we added them: reasonable assurance that independently generated ids will not clash. I realize that by using the shortIds at all, we sacrifice some of that, though for locally generated ones (i.e. inside publishing) we have do a uniqueness test. I also like that they are a standard type, that we can test for correctness at multiple levels in the system: e.g. the db has a uuid type, the BE server has a uuid type, the FE code has a uuid type: any and all of them can test it for correctness w/ code that someone else wrote. Lastly, it does reduce the cognitive load for an external developer to understand our data structures and content models and code: it's a known thing with known behavior, rather than an opaque id or some sort.

@pumazi as to checking each time, there are two occurrences of the use of split_ident_hash where we may not know the id in the code: one for the /contents route, one for the /extras route. These are the only times we process external input (a url from the FE). All other uses ignore the returned typeid. I suppose we could add some asserts to the code, to catch "something's wrong" cases. The  split_ident_hash _is_ the single place to handle this, in my mind. Since we can't convert from short to long without a DB connection, this is cleaner than passing a connection in to that method, same as the missing version issue.

cnxarchive/tests/test_utils.py
```diff
@@ -9,9 +9,13 @@
 import uuid
 import unittest
 
+from ..utils import CNXHash, IdentHashSyntaxError
+
 
 class SplitIdentTestCase(unittest.TestCase):
 
+    from ..utils import CNXHash
```
karenc: This line is redundant.

cnxarchive/utils/ident_hash.py
```diff
@@ -82,3 +81,186 @@ def join_ident_hash(id, version):
     if version:
         join_args.append(version)
     return HASH_CHAR.join(join_args)
+
+
+class CNXHash(uuid.UUID):
```
karenc: I find it kind of strange that we're subclassing `uuid.UUID` class.  What is the reason for this?

karenc:
From our discussion today, if I understand correctly, the short id is used for webview only.  Webview will rewrite urls to use short ids.  UUIDs will not be visible to users anymore but will continue to be used in code.

As for short id collisions, it will be caught by the unique index.

cnxarchive/utils/ident_hash.py
```diff
@@ -65,7 +64,7 @@ def split_ident_hash(ident_hash, split_version=False):
             if len(split_version) == 1:
                 split_version.append(None)
             version = tuple(split_version)
-    return id, version
+    return id, version, id_type
```
karenc: Since most callers don't care about the `id_type`, can we make it optional?

cnxarchive/scripts/hits_counter.py
```diff
@@ -93,14 +93,18 @@ def main(argv=None):
...
             for ident_hash, hit_count in hits.items():
-                cursor.execute(SQL_GET_MODULE_IDENT_BY_UUID_N_VERSION,
-                               split_ident_hash(ident_hash))
-                module_ident = cursor.fetchone()
-                payload = (module_ident, start_timestamp, end_timestamp,
-                           hit_count,)
-                cursor.execute("INSERT INTO document_hits "
-                               "  VALUES (%s, %s, %s, %s);",
-                               payload)
+                id, version, id_type = split_ident_hash(ident_hash)
```
karenc: If we make the `URL_PATTERN_TMPLT` stricter to match against uuids, we won't need to check the type.  (and I think we should only match the urls that use uuids.)

karenc:
Since we're trying to not import code (especially from cnx-archive) in postgres, I suggest that we remove the `base642uuid`, `uuid2base64` and identifier equality postgres functions unless we really need them.  We can already look up a module given a short id from the database.

I would also suggest that we move all the equal and similar methods in `CNXHash` to within the tests.  They can help verify that the generated short ids or base64 encoded uuids are indeed correct.  There's no need to have them in the code.

This is similar to what @pumazi has already said, shall we have more specific routes to match against uuids?  This way we know what kind of id we're expecting, we can have different handlers for short ids and I think the code might look a bit clearer?

cnxarchive/views.py
```diff
@@ -56,21 +56,36 @@ class ExportError(Exception):
...
+        try:
+            full_id = cursor.fetchone()[0]
+        except (TypeError, IndexError,):  # None returned
+            logger.debug("Short ID was supplied and could not discover UUID.")
+            raise httpexceptions.HTTPNotFound()
+    elif id_type == CNXHash.FULLUUID:
+        full_id = id
+    else:
+        logger.debug("Neither short_id nor full UUID: not implemented.")
+        raise NotImplemented()
```
karenc: Should this be `httpexceptions.HTTPNotFound()`?

reedstrm: yup, I agree.

reedstrm:
Agreed re: the extra methods in the SQL, and the test/equality methods on the class. If we need them in mainline code at some point in the future, they can be promoted from the tests. Perhaps Rich has ideas.

At first blush, it does seem that having separate routes for different uuid patterns would be cleaner.  Two things bother me about it, though: it requires encoding the logic for 'what does our ID look like' as regular expressions in the routes. While I read them like English :-), not everyone does: seems a lot more opaque to me than a function that takes a string and returns a CNXHash  - very centralized control of the data model. Secondly, I wonder if  we will need common code for the various cases of missing version, etc. Could be I'm wrong, though. Could be that the set of routes would be a clear way to describe the cases.

cnxarchive/utils/ident_hash.py
rich-hart: The UUID class is already used throughout our source code and it has its own datatype in postgresql.  Plus, the UUID class can already accept strings, hex digits and bytes.  Since I wanted to be able to calculate a short id no matter where a uuid was used or what form it was in.  Looking it over again, I would have liked to make it more 'python-y'.

cnxarchive/tests/test_utils.py
rich-hart: yeah, I'll remove it.

karenc:
We can give a name to the regular expression to make sure we know we're trying to match a uuid.  My current preferred way of doing it is still to have different routes.  We have common code for missing versions at the moment, but may be we can change that too.

cnxarchive/utils/ident_hash.py
karenc: What you said here is valid but doesn't mean we have to subclass `uuid.UUID`, we can always write a class that uses `uuid.UUID` (like storing a uuid object as an attribute).

The other thing I've noticed is that if we have so many class methods, we have to think about whether this should be a class.  What advantages do we have making this a class vs having functions?

reedstrm:
talked w/ Michael H a bit, he convinced me that web app developers would go looking for the routes first, rather than the python anyway. :-)

cnxarchive/utils/ident_hash.py
rich-hart: We could do that. The origenal reason I wanted to make a uuid.UUID subclass was so I could literally overwrite the class function with my own with an import line such as:

``` python
from utils import CNXHash as UUID
```

Then I was just going to invoke the get_shortid() and comparison functions where needed.  But the details of the implementation changed while working with Ross, so I acknowledge  that it doesn't make sense the way it is written now.  Let me see if i can refactor this some and maybe break the class in to separate functions.   The other reason why i grouped all this information in one class is that it's the easiest way to pull up the information that was used to calculate the short_id's from the python debugger.

reedstrm:
At scrum Friday, we decided to merge this and get it on dev "as is" to give FE devs @RoyEJohnson and @dak a chance to get the webview parts working. I'm going to _not_ delete the branch, and continue to work on the routing refactor I've been doing this week.

PR #367, opened 2015-10-27
add in-book-searching

mwharrison:


cnxarchive/sql/get-in-book-search.sql
```diff
@@ -0,0 +1,41 @@
...
+  tr.parent_id IS NULL
+UNION ALL
+SELECT c1.nodeid, c1.title, t.path || ARRAY[c1.nodeid], c1.documentid, t.depth+1, t.corder || ARRAY[c1.childorder]
+FROM trees c1 JOIN t ON (c1.parent_id = t.node)
+WHERE NOT nodeid = any (t.path)
+)
+SELECT
+M .uuid,
+M .version,
+COALESCE(T .title, M . NAME),
```
rich-hart: Just curious what the coalescing function is doing in this line?

mwharrison: It's getting the canonical form of the modules name.

cnxarchive/views.py
```diff
@@ -25,7 +25,7 @@
 from . import cache
 # FIXME double import
 from . import database
-from .database import SQL, get_tree
+from .database import SQL, get_tree, SQL_DIRECTORY
```
reedstrm: Oh, shouldn't be accessing the SQL_DIRECTORY from views.

cnxarchive/views.py
```diff
@@ -525,6 +525,137 @@ def get_export(request):
...
+    settings = get_current_registry().settings
+    connection_string = settings[config.CONNECTION_STRING]
+    with psycopg2.connect(connection_string) as db_connection:
+        with db_connection.cursor() as cursor:
+            if not version:
+                redirect_to_latest(cursor, id, 'in-book-search',
+                                   route_args=request.params.copy())
+
+            sql_file = os.path.join(SQL_DIRECTORY, 'get-in-book-search.sql')
+            with open(sql_file, 'r') as fp:
```
reedstrm: move the open/read code into database.py, so we only read it once on startup, not every search

cnxarchive/views.py
```diff
@@ -525,6 +525,137 @@ def get_export(request):
...
+    with psycopg2.connect(connection_string) as db_connection:
+        with db_connection.cursor() as cursor:
+            if not version:
+                redirect_to_latest(cursor, id,
+                                   'in_book_search_page',
+                                   route_args=request.params.copy())
+
+            sql_file = os.path.join(SQL_DIRECTORY,
+                                    'get-in-book-search-full-page.sql')
+            with open(sql_file, 'r') as fp:
```
reedstrm: ditto

cnxarchive/views.py
```diff
@@ -567,6 +698,10 @@ def search(request):
         page = None
     if page is None or page <= 0:
         page = 1
+    try:
+        in_book_uuid = params.get('uuid', '')
+    except (TypeError, ValueError, IndexError):
+        in_book_uuid = None
```
reedstrm: is this leftover from trying to overload the search method? Not sure what it's doing.

cnxarchive/views.py
```diff
@@ -525,6 +525,137 @@ def get_export(request):
...
+
+
+@view_config(route_name='in_book_search_page', request_method='GET')
+def in_book_search_highlighted_results(request):
+    """ in-book search - returns a highlighted version of the HTML """
+    results = {}
+
+    args = request.matchdict
+    ident_hash = args['ident_hash']
+    args['ident_hash'] = args['ident_hash'].split('@')[0]
```
reedstrm: this throws away the version

cnxarchive/views.py
```diff
@@ -525,6 +525,137 @@ def get_export(request):
...
+    args['page_ident_hash'] = args['page_ident_hash'].split('@')[0]
+
+    try:
+        args['search_term'] = request.params.get('q', '')
+    except (TypeError, ValueError, IndexError):
+        args['search_term'] = None
+
+    # Get version from URL params
+    id, version = split_ident_hash(ident_hash)
+    if version:
```
reedstrm: This will always be false, since you threw  the version away when parsing the ident_hash above, no?

mwharrison:
Good to know about the routes.. those were getting really hard to keep <80 chars. I'll update them to regain readability.

cnxarchive/views.py
mwharrison: The second line does, to return it in the json response. The first line 

`ident_hash = args['ident_hash']`

Preserves it for use in version parsing.
Now that you bring it to my attention, this is a little messy, let me see how I can refactor.

cnxarchive/views.py
mwharrison: Again, It won't be false, since I preserved it in the ident_hash variable; but it's not clear I see..

cnxarchive/views.py
mwharrison: It was! Removing!

cnxarchive/views.py
reedstrm: Ah I see my mistake. Carry on. (Clearer would be better, of course)

PR #370, opened 2015-11-10
legacy page versions

mwharrison:


cnxarchive/views.py
```diff
@@ -603,7 +603,7 @@ def in_book_search(request):
             for uuid, version, title, headline, rank in res:
                 results['results']['items'].append({
                     'rank': '{}'.format(rank),
-                    'id': '{}@{}'.format(uuid, version),
+                    'id': '{}@{}'.format(uuid, version.split('.')[1]),
```
pumazi: Stikes me as odd that we would be using the legacy version in archive. I'd think we'd only be using it explicitly or when it's absolutely necessary. Otherwise we should be using major and minor version with coalesce in sql or join_ident_hash in python.

reedstrm: Yeah, I spotted this and suggested @mwharrison fix it a different way. Push should be here in a minute.

PR #373, opened 2015-11-11
WIP: add activate_venv function

mwharrison:


mwharrison:
I found the method that is called when setting up the database, but it's still complaining the function does not exist.

pumazi:
Let me know when this is ready for review.

mwharrison:
As long as Travis passes, this is ready to go.

mwharrison:
@pumazi - You will also want to look at this gist: https://gist.github.com/mwharrison/384e8e64a6454b0e67ec

You need to have session_exec built on your local machine, if you don't already.

cnxarchive/database.py
```diff
@@ -112,6 +113,36 @@ def initdb(settings):
             for schema_part in get_schema():
                 cursor.execute(schema_part)
 
+            """If virtualenv is active, use that for postgres"""
```
pumazi: Comment rather than a string?

cnxarchive/database.py
```diff
@@ -112,6 +113,36 @@ def initdb(settings):
...
+            if hasattr(sys, 'real_prefix'):
+                import os
```
pumazi: Any reason we shouldn't move this to the top of the file?

cnxarchive/database.py
```diff
@@ -112,6 +113,36 @@ def initdb(settings):
...
+                                   "'session_exec'".format(db_name))
+                    cursor.execute("ALTER DATABASE \"{}\" SET "
+                                   "session_exec.login_name = "
+                                   "'venv.activate_venv'".format(db_name))
+                    cursor.execute("""CREATE FUNCTION venv.activate_venv()
+                                   RETURNS void LANGUAGE plpythonu AS $_$
+                                   import os
+                                   activate_this =
+                                   os.path.join('{}/bin/activate_this.py')
+                                   exec(open(activate_this).read(),
```
pumazi: Maybe use ``execfile`` instead. Or better yet, let's compile the file to bytecode so we don't read from the file system on each connection.

pumazi:
Can we start a pull request for @reedstrm's fork? The sooner we get that discussion started the better.

I'll try this as soon as I can.

cnxarchive/database.py
mwharrison: Good point - it's actually already up there! Double importing, tsk tsk.

cnxarchive/database.py
reedstrm: That would lock the initialized DB to the particular venv it was initialized from, then, right? So, effectively lock the trigger code and anything else imported to that venv.

pumazi: Right. That's the idea, no? It's locked to a specific virtualenv anyhow via the ``virtualenv_path`` variable.

cnxarchive/database.py
```diff
@@ -112,6 +113,34 @@ def initdb(settings):
...
+                                   "session_exec.login_name = "
+                                   "'venv.activate_venv'".format(db_name))
+                    cursor.execute("CREATE FUNCTION venv.activate_venv() "
+                                   "RETURNS void LANGUAGE plpythonu AS $_$\n"
+                                   "import os\n"
+                                   "activate_this = os.path.join('{"
+                                   "}/bin/activate_this.py')\n"
+                                   "exec(open(activate_this).read(),"
+                                   "dict(__file__=activate_this))"
+                                   "$_$".format(virtualenv_path))
```
karenc: I think if we use heredoc it might be easier to read:

```python
                         cursor.execute("""\
CREATE FUNCTION venv.activate_venv() RETURNS void LANGUAGE plpythonu AS $_$
import os
activate_this = os.path.join('{}/bin/activate_this.py')
exec(open(activate_this).read(), dict(__file__=activate_this))
$_$""".format(virtualenv_path))
```

karenc:
+1 for creating a pull request for @reedstrm's fork: https://github.com/okbob/session_exec/compare/master...reedstrm:master

cnxarchive/database.py
mwharrison: This is how I had it, @reedstrm thought this way was more readable. :smiley: 
I don't care either way, but to conform to PEP8, it will still need to wrap lines using heredoc style.

reedstrm:
This is going to need some runtime protections to not use different virtualenv from the one the db was inited with. In particular, it should also refuse to start against a remote DB that uses a venv. Sort of feels like we need a db schema version # somewhere that we test against on startup.

cnxarchive/database.py
```diff
@@ -112,6 +113,57 @@ def initdb(settings):
...
+sys.real_prefix = sys.prefix
+sys.prefix = base
+# Move the added items to the front of the path:
+new_sys_path = []
+for item in list(sys.path):
+    if item not in prev_sys_path:
+        new_sys_path.append(item)
+        sys.path.remove(item)
+sys.path[:0] = new_sys_path
+$_$""".format(activate_path=activate_path)
```
mwharrison: @pumazi @karenc @rich-hart  - Any thoughts on how we have basically ported the activate_this.py in here (to address a previous comment from @pumazi)?

pumazi: I don't know that I see anything wrong with porting this code here. I was originally suggesting you compile the activate_this module.

I believe that would look something like the following...

Before the sql function definition compile and marshal the module to bytecode.
```python
file = 'activate_this.py'
filepath = os.path.join('bin', file)
c = compile(open(file).read(), file, 'exec')
import marshal
bytecode = io.BytesIO()
marshal.dump(c, bytecode)
```
Then the bytecode is used to format the sql function, something like...
```python
filepath = "{filepath-formated-from-initdb}"
import marshal
bytecode = io.BytesIO("{bytecode-formated-from-initdb}")
c = marshal.load(bytecode)
exec(c, dict(__file__=filepath)
```

I don't know how well this will work without actually giving it a try. I think the ported version you came up with should work fine enough. Plus, the port will likely make it easier to read.

pumazi:
Can we make this whole segment that creates the active_venv sql function a separate function from initdb? It'll both compress the readability of initdb as well as (possibly?) make this bit reusable. The reusability may come in handy for if the virtualenv is ever moved.

cnxarchive/database.py
```diff
@@ -112,6 +113,60 @@ def initdb(settings):
...
+            # don't initalize with virtualenv
+            db_dict = dict(p.split('=') for p in db_connection.dsn.split())
+            if 'host' not in db_dict.keys() or ('host' in db_dict.keys() and
+                                                db_dict['host'] ==
+                                                'localhost'):
+                # If virtualenv is active, use that for postgres
+                if hasattr(sys, 'real_prefix'):
+                    activate_path = os.path.join(
+                        os.path.realpath(sys.prefix),
+                        'bin/activate_this.py')
```
rich-hart: Great!  Nice job auto linking to the virtualenv.

rich-hart:
I agree with @pumazi on making active_venv it's own utility function.  I would be great it could also be specified as its own script in the [cnx-archive/cnxarchive/scripts/](https://github.com/Connexions/cnx-archive/tree/master/cnxarchive/scripts) directory and adding it to our setup.py file [here](https://github.com/Connexions/cnx-archive/blob/master/setup.py#L45,L47).  That way maybe it could also be used to our virtualenv to the authoring database.

mwharrison:
Alright, I'll give that a try! Thanks.

cnxarchive/database.py
reedstrm: Right, we took a look at how much activate_this was doing, and it wasn't much: as you can see, the compile/marshal code is almost as many lines. I'm of the opinion we should remove some of the code, for dealing w/ PATH, since it's not likely to be needed/used from inside plpythonu.

mwharrison: Is this still true since the virtualenv paths needs to be put in front of the system paths?

cnxarchive/database.py
```diff
@@ -112,6 +113,60 @@ def initdb(settings):
             for schema_part in get_schema():
                 cursor.execute(schema_part)
 
+            # If you are connecting to a database that is not localhost,
+            # don't initalize with virtualenv
+            db_dict = dict(p.split('=') for p in db_connection.dsn.split())
+            if 'host' not in db_dict.keys() or ('host' in db_dict.keys() and
```
karenc: I think this can be simplified to `if db_dict.get('host', 'localhost') == 'localhost':`

cnxarchive/database.py
```diff
@@ -112,6 +113,60 @@ def initdb(settings):
             for schema_part in get_schema():
                 cursor.execute(schema_part)
 
+            # If you are connecting to a database that is not localhost,
```
karenc: May be the logic should be reversed.  Right now, it's checking whether the database is local and if it is, check whether we're using virtualenv.  

It makes more sense to me if we check whether we're using virtualenv first, and if the database is remote, raise an exception, if the database is local, set up virtualenv for postgres.  If we're not using virtualenv, whether the database is remote or not, there's nothing to do and no errors to raise.

cnxarchive/database.py
karenc: If we're going to port the code here, can we add a comment saying that this code is originally from virtualenv activate_this.py? (we might want to update the code if virtualenv changes)

If we're going to use marshal, I tried the code above and `marshal.dump` works with `io.BytesIO` in python 3.4 but not in python 2.7 (TypeError: marshal.dump() 2nd arg must be file).  So the updated code can be:

```python
file = 'activate_this.py'
filepath = os.path.join('bin', file)
c = compile(open(file).read(), file, 'exec')
import marshal
bytecode = marshal.dumps(c)
```

```python
filepath = "{filepath-formated-from-initdb}"
import marshal
bytecode = "{bytecode-formated-from-initdb}"
c = marshal.loads(bytecode)
exec(c, dict(__file__=filepath)
```

karenc:
It's good that we've found a way to load virtualenv in postgres, but I'm hoping eventually we won't have complicated python triggers when the legacy system is retired.

mwharrison:
The point of this branch is to make tests pass locally for developers. I think @reedstrm and myself agree that this has been accomplished and if the use cases expand, it should be revisited (to, for example, factor out a reusable function). 

@karenc - looks like a great refactor on the db connection string dictionary. I'll incorporate that.

karenc:
What about @reedstrm's pull request?

reedstrm:
My PR (against upstream session_exec) isn't really needed by any developers: we're all working on >=9.4. I _probably_ will make a PR to change the name of the function executed (from login_name to function_name, so the param becomes session_exec.function_name which I think reads nicely)

I agree w/ Karen that the test should be the other way around, but this is inside initdb, not run very often. It will catch the case that you pointed at the wrong ini file, basically.
It might be nice to make this a runtime check as well, but let's take this a step at a time, so factoring it out into a db helper method called by initdb makes some sense, I guess.

rich-hart:
runs great. merging.

karenc:
I feel like this was not ready to be merged yet since discussion hasn't finished yet.  What was the reason for merging it?  Our servers have been running fine without it so I didn't think there's any urgency to merge it right now. 

- Add a comment to say that we ported the code from virtualenv activate_this.py
- Test whether we're using virtualenv first
  - if the database is remote, raise an exception
  - if the database is local, create the activate venv function
- Factor the activate venv function creation code into a db helper function
- runtime check in activate venv

rich-hart:
@karenc my apologies.  I thought that we were about finished.  I thought that the remaining points in our discussion were mostly about features to be implemented later. Feel free to revert his branch if you don't think its ready yet.

rich-hart:
@karenc I was just reminded what was discussed in scrum this morning.  Our feeling during scrum is that it works for our dev purposes which is what is important. There isn't a urgency to merge it, but at the same time I don't feel there is any reason to delay merging it either.   I think that it would be good to keep it merged as it.  But I'm going to make an issue of your recommendations, because I think that they are good features that we should add.  That cool?

PR #383, opened 2015-12-04
Fix readme installation instructions to use session_exec

pumazi:
... documentation update

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -82,6 +82,24 @@ On a Debian based system, the installation is as follows::
     make && make install
     cd ..
 
+Installing the PostgresSQL session_exec extension
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is optional, but required if you choose to install the python packages
+into a virtualenv.
+
+::
+
+    git clone https://github.com/okbob/session_exec
+    cd session_exec
+    make USE_PGXS=1 && make USE_PGXS=1 install
+
+If you are using PostgreSQL 9.3 rather than >= 9.4, clone `reedstrm/session_exec <https://github.com/reedstrm/session_exec>`_ instead.
+
+This Postgres Extension is used to activate the virtualenv site-packages on
+any successful connection to the database, which then allows for importing
+packages that are only installed in the virtualenv.
+
 Set up the database and user
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -82,6 +82,21 @@ On a Debian based system, the installation is as follows::
...
     cd ..
 
+Installing the PostgresSQL session_exec extension
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is optional, but required if you choose to install the python packages
+into a virtualenv.
+
+::
+    git clone https://github.com/okbob/session_exec
```
karenc: So it's not necessary to use reedstrm's fork?

pumazi: My guess is that production is the only one using 9.3 at this point. Maybe I'm hinting at something. ;-)

karenc: I wasn't able to build session_exec for postgres 9.3 (what we were using on travis) because `session_preload_libraries` is not available for postgres 9.3...  Is it meant to work?

README.rst
```diff
@@ -82,6 +82,24 @@ On a Debian based system, the installation is as follows::
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is optional, but required if you choose to install the python packages
+into a virtualenv.
+
+::
+
+    git clone https://github.com/okbob/session_exec
+    cd session_exec
+    make USE_PGXS=1 && make USE_PGXS=1 install
```
karenc: I think `make USE_PGXS=1 install` requires `sudo`.

pumazi: It may, but that's for the user to determine. This is remaining consistent with the other blocks (e.g. plxslt).

README.rst
pumazi: You're right. This is why reedstrm's fork exists. I added a blob under the code block to note the version difference.

README.rst
karenc: You mean it works without `sudo`, so it's for the user to determine whether to use `sudo` or not?

pumazi: Depending on how your system is setup and which user you are operating as, Yes.

reedstrm: A typical linux install will need sudo. A typical OSX developer install (using postgresql.app), perhaps not.

pumazi: Ok, sudo is probably the way to go here, but it's not within the scope of this pull request. I was just fixing the missing documentation. If you guys want sudo, make another pull request for it. :pineapple:

PR #391, opened 2016-01-07
Move short id equality and similarity functions to tests

karenc:
- Move the content of identifiers_equal.sql into test_database.py

  The code has no use case at the moment, so going to move to the tests
for now.

- Type cast args in identifiers_equal and use the one that accepts text

  
- Move identifiers equal code from CNXHash to test utils

  It is not possible to import identifiers_equal from cnxarchive.tests in
plpython (test_database.py), so just copy the source code into the sql
function. :p

- Move identifiers similar code to test utils

  
- Remove base642uuid sql function

  Not actually used at the moment.

- Remove uuid2base64 sql function

  Not actually used at the moment.


reedstrm:
:+1:

PR #393, opened 2016-01-07
documentation for content API

edwoodward:


philschatz:
:+1: :sunglasses:

philschatz:
Some other questions:

- Should it include the syntax for optionally including the version? 
- Which parts of the extra stuff in the webview URLs work/dont? ie `...[BOOK_UUID]@123:4/Physics_Title` (I think people might wonder why things look broken when they just replace `cnx.org` with `archive.cnx.org`)
  - does including the `[PAGE_UUID]` work?
  - Does the `/Physics_Title` part work?
- Should we list the extras URLs for getting PDFs/Zips, and the sims?

edwoodward:
I was trying to document what was needed by WebAssign. They are only interested in HTML and the latest version as far as I know. I agree I did not document everything.

edwoodward:
Added notes about using short ids and version numbers

reedstrm:
ShortIds work just fine - as Phil mentions, it's the extra title part that's not currently ignored.

philschatz:
ah, I didn't know who the audience was. Could/should there be a link to this doc from the `README.rst`?

reedstrm:
And I think this page should seek to document the whole API, we can subset if need be for other audiences.

reedstrm:
BTW, Karen's is refactoring this code now,  - it's somewhat likely that this will regularize things like ignoring the extra title cruft

pumazi:
Ideally this documentation would be in the doc strings of the various views and pulled into a restructuredtext document. I think this works for the time being.

edwoodward:
I tested the short id for physics and got a 404. It worked on a second try. I'll update the doc.

reedstrm:
@pumazi - that sounds cool, though there's the disconnect between the route-name (which is what the view knows about) and the actual route, as defined by the RE in the add_route call. I'd hate to propagate examples URLs in docstrings that are in a different source file than the code that defines them. (though that's what's happening here as well, given that it's only doco, it's understood that the code is elsewhere)

reedstrm:
BTW, I was talking out of my hat as well - we already ignore all the extra URL cruft.

content_api_doc.md
```diff
@@ -0,0 +1,42 @@
+Content API
```
karenc: How about putting this (and all the other documentation that's not README) into a `docs/` directory?  As @philschatz said, we can link to it in README.

content_api_doc.md
```diff
@@ -0,0 +1,42 @@
...
+ * In the browser, a short id is displayed for books and pages. There is also a longer UUID for books and pages. Either id will work in the API. The full UUID for a Book and the current Page can be found in the More Information tab at the bottom of any content.
+ * The title should not be included in the content API. request.
+
+**Book metadata and Table of Contents as json**
+
+  >GET http://archive.cnx.org/contents/[book ID].json
+
+Examples (long UUID and short id with version number)
+
+  >http://archive.cnx.org/contents/031da8d3-b525-429c-80cf-6c8ed997733a.json
```
karenc: Doesn't need to be in this PR but what do people think about turning this into a doctest? ;)

reedstrm: :+1:  to both - do it, but merge this first.

content_api_doc.md
reedstrm: oh, agree to this as well. :+1:

README.rst
```diff
@@ -210,6 +210,11 @@ Then, use either of the following to invoke the test suite::
 
 This uses example data found in the ``cxarchive/tests/data`` directory.
 
+Usage
+-----
+ * [Content API](/docs/content_api_doc.rm)
```
philschatz: rename to `.md`?

pumazi: The rst syntax you are looking for is `` `Content API <docs/search_api_doc.md>`_ ``

philschatz: yep got it. thanks!

PR #394, opened 2016-01-18
Bug fixes for in book search

karenc:
- Simplify in book search version code

- Fix in book search redirect to latest code

- Fix in book page search redirect to latest code

- Catch IdentHashSyntaxError for all the views

- Add coveralls badge to README

---

Route refactoring will be in another PR

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,9 @@ Running tests
 .. image:: https://travis-ci.org/Connexions/cnx-archive.png?branch=master
    :target: https://travis-ci.org/Connexions/cnx-archive
 
+.. image:: https://coveralls.io/repos/Connexions/cnx-archive/badge.svg?branch=master&service=github
+   :target: https://coveralls.io/github/Connexions/cnx-archive?branch=master
+
 The tests use the standard library ``unittest`` package and can therefore
 be run with minimal effort. Make a testing config, such as testing.ini,
 and set the environment variable ``TESTING_CONFIG`` to the name of that file::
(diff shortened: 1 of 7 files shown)


pumazi:
Rather than use a decorator for the ident syntax error, let's use a [custom exception view](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/views.html#custom-exception-views). The decorator you've implemented looks more like a flask solution, which is fine if your into that kinda thing. ;)

pumazi:
Hmm... What do you think of using the same custom exception view approach for a request with a shortid? Have ident_hash parser raise a ``ShortID(ident_hash)`` (probably a better name for this), then have the custom exception view for ShortID redirect to the canonical.

I'm not a huge fan of stacking decorators, so getting away from stacked decorators as much as possible is a win in my book. :see_no_evil:

karenc:
Oh, I'm unfamiliar with flask, decorator was just a pure python solution. I'll use the exception view for catching ident syntax error. Which is much better because we won't need to add it per view. :D

karenc:
What do you think about using more specific regexps in routes? I prefer it but kind of hesitated because a lot more code will need to be changed. :p

karenc:
Changing this PR to just do bug fixes.  Route refactoring will be in another PR.

reedstrm:
re: more speciifc regex's - I'm all in favor of the deeper refactor.

karenc:
If this looks ok, want to merge this first?

reedstrm:
This fix is more complete than mine in #395, so I'll merge this and rebase that to be just the book html bit.

reedstrm:
@karen you've got the other code elsewhere, so I should delete this branch, or should I just leave it?

karenc:
@reedstrm You can delete the branch.  I'll push a new one.

PR #395, opened 2016-01-19
Html tree depth

reedstrm:
Mixed PR, I'm afraid. Alter html form of a book to expose the depth-of-nesting (in lieu of chapter/unit/etc.) Also, add some tests to in-book-search.

### DIFF
--- a/cnxarchive/tests/test_views.py
+++ b/cnxarchive/tests/test_views.py
@@ -972,23 +972,23 @@ class ViewsTestCase(unittest.TestCase):
 
         expected = u"""<html xmlns="http://www.w3.org/1999/xhtml">
   <body>\
-<ul><li><a href="/contents/e79ffde3-7fb4-4af3-9ec8-df648b391597%407.1.html">College Physics</a>\
-<ul><li><a href="/contents/209deb1f-1a46-4369-9e0d-18674cf58a3e%407.html">Preface</a></li>\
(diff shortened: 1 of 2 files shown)


karenc:
I got more in-book-search fixes in the first 4 commits of #394.  Maybe I'll remove the other commits so that can be merged.

reedstrm:
I'll rebase this to just the html part.

pumazi:
What's the basis for needing the ``data-depth`` attribute on these tags?

Seems like you could get this info in CSS and XPath without explicitly defining it. Also, this html isn't used anywhere except in the archive html view. Are you using this somewhere else?

pumazi:
Now that cnx-archive depends on cnx-epub we can move this logic there. There is actually a copy of these functions within the adapters module.

reedstrm:
As we discussed in a meeting, this change is on the path toward assembling a nested-html to represent a full book (nodes _and_ leaves). That's probably going to take other code, since we can't just include the page html files (can't nest <html> or <body> tags), so some processing will be needed.
I'm still thinking that it's within archive portfolio to provide it. Perhaps publish should push a 'div-ified' version of each page as well, for this use case.

BTW,  depth-of-nesting is actually somewhat tricky to do with CSS or XPATH. Full-on XSLT has a count() function that could serve, but we're trying to remove XSLT from the pipeline.

Closing, as not currently in scope.

PR #396, opened 2016-01-21
More in book search fixes (and one other fix ...)

reedstrm:
shortId w/ in-book-search, pep8, and make the schema match deployed triggers (explicit column names, not *)


karenc:
Do you mind if I redo the commits?

pumazi:
:+1: :shipit:

PR #397, opened 2016-01-22
Refactor routes

karenc:
Continuation of #394

- Refactor get uuid from shortid code and get latest version code

  
- Change redirect_to_canonical to use matched route from request

  
- Get route name/args and params from request in redirect to canonical

  
- Remove duplicate test from ViewsTestCase

  test_truncate_hash is the same as test_content_shortid_version.

- Simplify is_latest helper function in views

  
- Move view helper functions to the top of the file

  
- Remove some unnecessary try-except blocks from views.py

- Remove unnecessary line "len(res)" from views in_book_search

- Replace redirect to canonical with exception views

  Instead of checking for missing versions and short ids, raise exceptions
IdentHashMissingVersion and IdentHashShortId when splitting ident hash
and redirect in exception views.

- Remove return_type from split_ident_hash

  split_ident_hash is raising IdentHashShortId for short ids, so there
is no need to return the type anymore.

- Add IdentHashError base exception class for all ident hash exceptions

### DIFF
--- a/cnxarchive/utils/ident_hash.py
+++ b/cnxarchive/utils/ident_hash.py
@@ -50,7 +50,7 @@ def split_legacy_hash(legacy_hash):
     return id, version
 
 
-def split_ident_hash(ident_hash, split_version=False, return_type=False):
+def split_ident_hash(ident_hash, split_version=False):
     """Returns a valid id and version from the <id>@<version> hash syntax."""
     if HASH_CHAR not in ident_hash:
         ident_hash = '{}@'.format(ident_hash)
@@ -78,10 +78,7 @@ def split_ident_hash(ident_hash, split_version=False, return_type=False):
         if len(split_version) == 1:
             split_version.append(None)
         version = tuple(split_version)
-    if return_type:
-        return id, version, id_type
-    else:
-        return id, version
+    return id, version
 
 
 def join_ident_hash(id, version):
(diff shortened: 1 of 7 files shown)


karenc:
Ready for review.

cnxarchive/utils/ident_hash.py
```diff
@@ -25,6 +27,19 @@ class IdentHashSyntaxError(Exception):
...
 
 
+class IdentHashShortId(Exception):
+    """Raised when the ident-hash id is not a uuid"""
+    def __init__(self, id, version):
+        self.id = id
+        self.version = version
+
+
+class IdentHashMissingVersion(Exception):
```
pumazi: Let's put an IdentHashError base class in place. This way we can except on IdentHashError for all ident-hash problems.

cnxarchive/utils/ident_hash.py
```diff
@@ -35,7 +50,7 @@ def split_legacy_hash(legacy_hash):
     return id, version
 
 
-def split_ident_hash(ident_hash, split_version=False, return_type=False):
+def split_ident_hash(ident_hash, split_version=False):
```
pumazi: This is much better! :+1:

cnxarchive/utils/ident_hash.py
karenc: Good idea.  Added IdentHashError.

PR #399, opened 2016-01-26
Remove plpydbapi

karenc:
- Rewrite module and collection republish triggers without plpydbapi

- Rewrite transforms without plpydbapi

- Remove dependency plpydbapi


pumazi:
cnx-publishing tests pass as well.

PR #402, opened 2016-01-28
Add a list of resources to the content response.

pumazi:
This provides a means of acquiring the cover image and RuleSet
associated with a book.


pumazi:
Any thoughts on possibly adding a meta-type to the ``module_files`` table? So, meta-types might be one of ``( 'content', 'resource', 'cover', 'ruleset' )`` to indicate the purpose of the file entry.

This would allow us to potentially do a more narrow select on anything except 'content'. This isn't necessary since it doesn't really hurt to just display everything. IMO it would help better define the data, but it's not necessary.

This would also allow for resource discovery by folks using the API. So the output for the collection might look something like:

```
{ ...
"resources" : [
        {"filename": "featured-cover.png",
         "type": "cover",
         "id": "6214e8dcdf2824dbf830b4a0d77a3fa2f53608d2",
         "mimetype": "image/png"
         },
        {"filename": "collection.html",
         "type": "content",
         "id": "921dcf515d41c5a5cbe3c2163ed0f5db02ab3c83",
         "mimetype": "text/html"
         },
        {"filename": "collection.xml",
         "type": "content",
         "id": "921dcf515d41c5a5cbe3c2163ed0f5db02ab3c83",
         "mimetype": "text/xml"
         }
    ]
...
}
```

karenc:
I think adding a meta-type to `module_files` is good.

reedstrm:
Since we're discussing schema changes:
but the mime-type really shouldn't be in the modulefiles table, it should be in the files table, since any given bit sequence can only have one.

reedstrm:
Should we consider suppressing (or marking) for modules the 'resource' that is represented by the contents node of the same json?

PR #405, opened 2016-02-16
Raise ident hash syntax error if ident hash ends with @

karenc:
Change ident hash syntax error exception view to try again with the
ident hash without the @ sign.

### DIFF
--- a/cnxarchive/utils/ident_hash.py
+++ b/cnxarchive/utils/ident_hash.py
@@ -25,12 +25,15 @@ __all__ = (
 
 class IdentHashError(Exception):
     """Base exception class for all ident hash exceptions."""
+    def __str__(self):
+        return self.msg
 
 
 class IdentHashSyntaxError(IdentHashError):
     """Raised when the ident-hash syntax is incorrect."""
     def __init__(self, ident_hash):
         self.ident_hash = ident_hash
+        self.msg = 'ident_hash={}'.format(ident_hash)
 
 
 class IdentHashShortId(IdentHashError):
@@ -38,12 +41,14 @@ class IdentHashShortId(IdentHashError):
     def __init__(self, id, version):
         self.id = id
         self.version = version
+        self.msg = 'id={} version={}'.format(id, version)
 
 
 class IdentHashMissingVersion(IdentHashError):
     """Raised when the ident-hash does not have a version"""
     def __init__(self, id):
         self.id = id
+        self.msg = 'id={}'.format(id)
 
 
 def split_legacy_hash(legacy_hash):
--- a/cnxarchive/utils/ident_hash.py
+++ b/cnxarchive/utils/ident_hash.py
@@ -29,6 +29,8 @@ class IdentHashError(Exception):
 
 class IdentHashSyntaxError(IdentHashError):
     """Raised when the ident-hash syntax is incorrect."""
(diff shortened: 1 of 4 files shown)


cnxarchive/utils/ident_hash.py
```diff
@@ -56,19 +58,17 @@ def split_legacy_hash(legacy_hash):
...
 
-    try:
-        id, version = split_value
-    except ValueError:
+    if len(split_value) > 2 or '' in split_value:
         raise IdentHashSyntaxError(ident_hash)
 
-    # Validate the id.
+    if len(split_value) == 1:
+        id, version = split_value[0], ''
```
pumazi: Is it possible to raise ``IdentHashMissingVersion `` here? Or is that too soon?

karenc: This is a question of whether to raise `IdentHashShortId` first or `IdentHashMissingVersion` :stuck_out_tongue:

karenc: I think we raise `IdentHashShortId` first so we can get the uuid and then if the version is also missing, we can get the version from the uuid. :smile:

cnxarchive/utils/ident_hash.py
```diff
@@ -29,6 +29,8 @@ class IdentHashError(Exception):
 
 class IdentHashSyntaxError(IdentHashError):
     """Raised when the ident-hash syntax is incorrect."""
+    def __init__(self, ident_hash):
+        self.ident_hash = ident_hash
```
pumazi: The ``ident_hash`` needs added to ``self.args``, otherwise It won't end up in the traceback. I know this because I've also made this change in an uncommited branch. :stuck_out_tongue: (see also, [the BaseException interface](https://docs.python.org/3.6/library/exceptions.html#BaseException))

cnxarchive/utils/ident_hash.py
pumazi: Yep, that sounds right. :taxi:

cnxarchive/utils/ident_hash.py
pumazi: Hrm... I was just looking at the other classes that subclass ``IdentHashError``. They suffer the same issue. Can you fix those too? :pray:

karenc: I've used `__str__` instead of `self.args`... :stuck_out_tongue:   I'll change it if it's not a good idea.  Exceptions will now look like this:

```
>>> raise IdentHashMissingVersion('8ac77e82-95d8-4a90-8b06-80a83bb8e636')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
cnxarchive.utils.ident_hash.IdentHashMissingVersion: id=8ac77e82-95d8-4a90-8b06-80a83bb8e636
>>> raise IdentHashShortId('12345678', 1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
cnxarchive.utils.ident_hash.IdentHashShortId: id=12345678 version=1
>>> raise IdentHashSyntaxError('a@')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
cnxarchive.utils.ident_hash.IdentHashSyntaxError: ident_hash=a@
```

reedstrm:
+1

PR #408, opened 2016-02-25
improve triggers on files table

reedstrm:
firing the triggers every time slows the migration phenomenally. Made them conditional on the file attribute changing. Updating all files media_types took 39 sec w/ selective triggers, 35 sec with no triggers, and never completed with existing triggers (interrupted at 17 minutes)


karenc:
:+1:

pumazi:
Why not disable the triggers on the table rather than drop them?

pumazi:
The other migration also uses the same ``update`` statement.

karenc:
To answer the why not disable trigger question, I think this is the trigger migration (bundled in your migration).

reedstrm:
@pumazi  two reasons: it's not trivial to disable triggers, and two, the new trigger is installed slighlty differently: only fires on update of the file attribute. That's also why it's not needed in the second migration: the update works fine in the presence of the limited trigger.  Hmm, seems the second migration should only update missing media types. I'll add that.

PR #409, opened 2016-02-26
Internal epub

pumazi:
This includes a means to create an [internal] EPUB from the commandline and HTTP interfaces.

This is by no means perfect, but it'll get us closer to having a complete replacement for the ``completezip`` format.

The reason for the various functions is because it's a step in a direction I hope we will eventually be able to merge into out other code. ...which will eliminate a lot of the SQL knowledge from the views layer.

The view is really slow... But this isn't suppose to be optimal. Plus these are to be generated on publishing anyhow. So sometime in the future, we'll be pulling them off the filesystem.


karenc:
:+1: Nicely structured commits and code plus lots of tests.  :100:

PR #410, opened 2016-02-26
Fix SQL syntax error in migrations

pumazi:
Simple fix for a SQL syntax error introduced in #408. Sorry, this is my fault for not catching it before I pushed the merge button. :confounded:


karenc:
Oh... Totally missed it. Thanks for fixing it :+1:

PR #411, opened 2016-02-26
Move the file trigger update to a separate migration

karenc:
It makes things easier because migrations that are already run on a
server should not really be changed.

### DIFF
--- a/cnxarchive/sql/migrations/20160101000000_update_file_trigger.py
+++ b/cnxarchive/sql/migrations/20160101000000_update_file_trigger.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+"""\
+Change file hash update triggers to only trigger on file update
+"""
+
+
+def up(cursor):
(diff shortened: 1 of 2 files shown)


reedstrm:
to reify my comment from the meeting - I think this is true, that migrations that have been run _on a production server_ should never be changed. But dev doesn't count (qa probably not, even)

karenc:
I disagree. I think unless a migration is wrong, we should just leave it alone.

Dantemss:
It all depends how important the data in dev is... if you are fine with totally dropping the dev DB then migrations that ran in dev don't matter :)

reedstrm:
Not being able to complete on production in a reasonable amount of time makes it wrong, in my book. That's the ops half of devops. That's also why this PR splitting the trigger changes out to a separate migration is exactly right - dropping/adding triggers is _much_ riskier in a production environment - depending on the load, it can stall essentially forever.

PR #416, opened 2016-03-14
Unique files table entries

pumazi:
This will deduplication files table entries and prevent them from happening in the future. This also contains a change to move the pep8 checks to the top of the travis config so that it fails fast.

For example:
```
                   sha1                   |  c
------------------------------------------+-----
 075500ad9f71890a85fe3f7a4137ac08e2b7907c | 186
 e0a621ecbb46ba16a9235263be5233eb0e6897c3 | 158
 40f706929ab55ad5d247f69b7b0918ddb9c49c58 | 137
 fff772d27029cb8e720c8be9db336c968f835d22 | 108
 4efceebae7d0d4e155a8ef775667a984ca76cbff | 101
 4e3c963bada0fbd8c67016a375bb44550a5ae295 |  71
 6e272a9430637f03e127ab214497cf3ba0666a3d |  71
 64690f3ebc5dd10a7c4718cbd4d2f1113b0155a6 |  65
...
```
Where `c` is the duplication count.

Note, there is no rollback for this. I could have put one in, but frankly it's way more work than it's worth. It may be faster to restore from backup if things went horribly wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ addons:
 services:
   - memcached
 before_install:
+  - pip install pep8
+  - pep8 --exclude=tests *.py cnxarchive/
+  - pep8 --max-line-length=200 cnxarchive/tests
+
   - sudo apt-get update
   # Remove python-zope.interface, not compatible with the pyramid version we're
   # using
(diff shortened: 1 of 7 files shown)


pumazi:
... and I'm still working on the companion pull request for cnx-publishing. :hourglass:

karenc:
I checked [the sql files in RhaptosModuleStorage](https://github.com/Rhaptos/Products.RhaptosModuleStorage/tree/084bc8158fca3c6c8b52617f5af44a8a583f5314/Products/RhaptosModuleStorage/sql) and from what I see, I don't think reading will be a problem, [inserting files](https://github.com/Rhaptos/Products.RhaptosModuleStorage/blob/084bc8158fca3c6c8b52617f5af44a8a583f5314/Products/RhaptosModuleStorage/ModuleDBTool.py#L224) looks ok too.

reedstrm:
legacy/rhaptos already de-dups - It retrieves first, only inserts if the hash does not exist, which I see is the exact logic this PR implements as well. Makes sense to me.

PR #423, opened 2016-04-11
Collation

pumazi:
Provides the schema fixtures for collation data. The current level of functionality is maintained while these changes are put in place. In other words, these changes do not directly effect behavior of the archive api.

This introduces a new column on the trees table and a new table for associating collated content.


karenc:
Looks good to me.

PR #424, opened 2016-04-11
Collated content

karenc:


reedstrm:
:+1:

cnxarchive/sql/search/query.sql
```diff
@@ -64,6 +64,7 @@ FROM
   derived_weighted_query_results AS wqr
 WHERE
   wqr.module_ident = lm.module_ident
+  AND lm.portal_type != 'CompositeModule'
```
pumazi: I guess we can't really search composite module content at the moment, but is there any reason not to search its metadata? Perhaps that's the problem, that the query breaks in the full text search part.

Perhaps we should note this failing somewhere?

cnxarchive/tests/data/data.sql
```diff
@@ -557,6 +572,7 @@ INSERT INTO module_files VALUES (4, 23, 'PhET_Icon.png');
 INSERT INTO module_files VALUES (6, 23, 'PhET_Icon.png');
 INSERT INTO module_files VALUES (9, 23, 'PhET_Icon.png');
 INSERT INTO module_files VALUES (10, 23, 'PhET_Icon.png');
+INSERT INTO module_files VALUES (20, 109, 'index.cnxml.html');
```
pumazi: Should this entry exist? From my perspective, `modules` records that are of `portal_type = 'CompositeModule'` should not have any `module_files` entries.

pumazi: Another way to look at this is... At this time `modules` records that are of type `portal_type = 'Module'` will not have a `module_files` entry for the collated content. That's why I'm questioning whether this CompositeModule entry should exist.

If we were to add the collated content entries to `module_files`, it should maybe have a new filename?

cnxarchive/utils/mimetype.py
```diff
@@ -23,6 +23,7 @@
 MIMETYPES = (MODULE_MIMETYPE, COLLECTION_MIMETYPE, FOLDER_MIMETYPE,)
 PORTALTYPE_TO_MIMETYPE_MAPPING = {
     'Module': MODULE_MIMETYPE,
+    'CompositeModule': MODULE_MIMETYPE,
```
pumazi: Would it be a good idea to give the CompositeModule a bona fide media-type?

cnxarchive/views.py
```diff
@@ -886,6 +897,7 @@ def sitemap(request):
                             '[^0-9a-z]+', ' ', 'g')), ' +', '-', 'g'),
                         revised
                     FROM latest_modules
+                    WHERE portal_type != 'CompositeModule'
```
pumazi: Any reason not to put the CompositeModule in the sitemap?

Perhaps we should address this later? If so, let's note it somewhere.

cnxarchive/utils/mimetype.py
reedstrm: Hmm, what will legacy do with those, then?

cnxarchive/tests/data/data.sql
karenc: Should it be in `collated_file_associations`?

cnxarchive/utils/mimetype.py
karenc: Will it be a problem in legacy to have `portal_type = 'CompositeModule'` in the modules table?

cnxarchive/views.py
karenc: @reedstrm What do you think about putting composite modules in the sitemap?

cnxarchive/tests/data/data.sql
karenc: Do composite modules only exist in the context of a book?

cnxarchive/utils/mimetype.py
karenc: Added media type for composite modules.

cnxarchive/tests/data/data.sql
pumazi: > Should it be in `collated_file_associations`?

Yes it should be part of that table, `<collection id>, 20, 109`

>  Do composite modules only exist in the context of a book?

Nope.

karenc: So how do we look up the content without knowing which collection it is in?

cnxarchive/sql/search/query.sql
karenc: Raised an issue: https://github.com/Connexions/cnx-archive/issues/425

cnxarchive/views.py
karenc: Raised an issue: https://github.com/Connexions/cnx-archive/issues/426

cnxarchive/tests/data/data.sql
pumazi: A composite module is singularly owned by a collection. So it cannot be acquired outside the context of the collection.

If we really really had to provide a route to composite modules, I think we could. Given the singular nature of the content, we could assume that any request for a composite module could be queried against `collated_file_associations` in conjunction with `latest_modules` if the version is missing. This would find the unique entry for the content.

cnxarchive/views.py
karenc: We're excluding composite modules from the sitemap for now.

cnxarchive/tests/data/data.sql
karenc: A composite module returns 404 if it is requested outside of a collection.  See db1f9b8

karenc:
Ready for review again. :)

cnxarchive/tests/data/data.sql
reedstrm: right, the only retrieval path for compositemodule should be via the contextual url, which will use the other table entry.

PR #427, opened 2016-04-22
Remove warning info from the migration

pumazi:
This select statement takes ~3hrs on a production machine.

### DIFF
--- a/cnxarchive/sql/migrations/20160128111115_mimetype_removal_from_module_files.py
+++ b/cnxarchive/sql/migrations/20160128111115_mimetype_removal_from_module_files.py
@@ -14,16 +14,6 @@ def up(cursor):
                    "FROM module_files AS mf "
                    "WHERE mf.fileid = f.fileid AND f.media_type IS NULL")
 
-    # Warn about missing mimetype.
-    cursor.execute("SELECT fileid, sha1 "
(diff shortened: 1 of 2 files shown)


cnxarchive/sql/migrations/20160128111115_mimetype_removal_from_module_files.py
```diff
@@ -14,19 +14,6 @@ def up(cursor):
...
-                   "FROM files AS f "
-                   "WHERE f.fileid NOT IN (SELECT fileid FROM module_files)")
-    rows = '\n'.join(['{}, {}'.format(fid, sha1)
-                      for fid, sha1 in cursor.fetchall()])
-    print("These files (fileid, sha1) do not have a corresponding "
-          "module_files entry:\n{}\n".format(rows),
-          file=sys.stderr)
-
-    # Remove the ``mimetype`` column from the ``module_files`` table.
-    cursor.execute("ALTER TABLE module_files DROP COLUMN mimetype")
```
karenc: We should still drop the mimetype column right?

pumazi: Oops! I guess, I got a bit delete happy. Good catch! 💯

PR #429, opened 2016-04-23
Add SQL functions for debugging plpythonu

pumazi:


karenc:
@pumazi Could you have a look at the commits I added to fix the broken test? :)  If they look ok, I'll merge this.

```
======================================================================
FAIL: test_pyimport (cnxarchive.tests.test_database.MiscellaneousFunctionsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/Connexions/cnx-archive/cnxarchive/tests/testing.py", line 74, in wrapped
    return method(self, cursor, *args, **kwargs)
  File "/home/travis/build/Connexions/cnx-archive/cnxarchive/tests/test_database.py", line 119, in test_pyimport
    self.assertEqual(directory, expected_directory)
AssertionError: '/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/cnx_archive-2.4.1-py2.7.egg/cnxarchive' != '/home/travis/build/Connexions/cnx-archive/cnxarchive'
----------------------------------------------------------------------
```

pumazi:
The problem with the use of `setup.py develop` is that it will no longer test for the installation of misc files. Or put another way, it won't test the MANIFEST.in.

karenc:
Ok, could you try to fix it in a different way?

pumazi:
I've tried a few things. And there doesn't appear to be a way around this short of writing our own test runner. This may be one of the assortment of reasons zope wrote it's own test runner. Basically, `setup.py test` puts the relative source in its path, `-m unittest discover -s $proj_dir` makes the project root the CWD and puts it as the relative source in python's path. So we are left with test runners that don't seem to want to run outside the root of the project. This seems very unfortunate.

I don't know that we should sacrifice the MANIFEST.in checks for the sake of this debug func. Maybe we should comment it out and potentially revisit this issue in the future. Maybe tools will get better in the near future? (Though, I say that skeptically, because python testing tools haven't changed for many years.)

karenc:
Yes, that's ok to me.  Perhaps we can change the test a bit to pass with less strict asserts.

Could you also remove the last 2 commits that I added?

karenc:
@pumazi I got it to work without making the tests less specific. :D  Could you have a look at the last commit?

PR #434, opened 2016-05-17
Added query to get a list of available languages in modules

hackthehackerman:


reedstrm:
How about we extend that a bit, to count the kinds of content, as is done for subjects? And especially with counts, make this against the latest_modules table - the modules table has every version of everything that's ever been published.

https://github.com/Connexions/cnx-archive/blob/master/cnxarchive/sql/get-subject-list.sql

Also, since this is just the SQL, mark the PR as [WIP]. The full PR should contain the python to use this under the /extras route.

hackthehackerman:
okay.

edwoodward:
Do you want the counts for future use? They will not be used on the Advanced Search page.

reedstrm:
@ed yes

PR #435, opened 2016-05-17
Query available languages

hackthehackerman:
Added a call that query available languages and their counts.


hackthehackerman:
I will resolve the test errors before creating another pull request.

PR #437, opened 2016-05-18
Query available languages

hackthehackerman:
Added a query to get a list of available languages for latest_modules and the count for each languages.

### DIFF
--- a/cnxarchive/views.py
+++ b/cnxarchive/views.py
@@ -399,6 +399,12 @@ def _get_subject_list(cursor):
         yield subject
 
 
+def _get_available_languages_and_count(cursor):
+    """Return a list of available language and its count"""
+    cursor.execute(SQL['get-available-languages-and-count'])
+    return cursor.fetchall()
+
+
 def _get_featured_links(cursor):
     """Return featured books for the front page."""
     cursor.execute(SQL['get-featured-links'])
@@ -881,6 +887,8 @@ def extras(request):
     with psycopg2.connect(settings[config.CONNECTION_STRING]) as db_connection:
         with db_connection.cursor() as cursor:
             metadata = {
+                'languages_and_count': _get_available_languages_and_count(
+                    cursor),
                 'subjects': list(_get_subject_list(cursor)),
                 'featuredLinks': _get_featured_links(cursor),
                 'messages': _get_service_state_messages(cursor),
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -41,6 +41,8 @@ def _read_sql_file(name):
         return fp.read()
 
 SQL = {
+    'get-available-languages-and-count': _read_sql_file(
+        'get-available-languages-and-count'),
     'get-module': _read_sql_file('get-module'),
     'get-content-from-legacy-id': _read_sql_file('get-content-from-legacy-id'),
     'get-content-from-legacy-id-ver': _read_sql_file(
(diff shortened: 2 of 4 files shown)


reedstrm:
This looks good, setting up the backend retrieval. We typically merge complete features to master, however. Let's get the changes to the extras view in there  (should be a one-liner right about https://github.com/Connexions/cnx-archive/blob/master/cnxarchive/views.py#L883-L888) and update the tests, and we'll be good to go.

hackthehackerman:
@reedstrm Sorry, I was not familiarized with the architecture. Based on the example, I assume I should add `'languages_and_counts': _get_available_languages_and_count(x),` to where you were pointing to?

reedstrm:
yup, looks like it should work. Then run 'python setup.py test' in the top level (cnx-archive) directory. I presume some tests will fail. Update the expected output.

cnxarchive/views.py
```diff
@@ -399,6 +399,12 @@ def _get_subject_list(cursor):
         yield subject
 
 
+def _get_available_languages_and_count(cursor):
+    """Return a list of available language and its count"""
+    cursor.execute(SQL['get-available-languages-and-count'])
+    return [i[0] for i in cursor.fetchall()]
```
karenc: Is this going to return just the language without the count?

cnxarchive/views.py
```diff
@@ -402,7 +402,7 @@ def _get_subject_list(cursor):
 def _get_available_languages_and_count(cursor):
     """Return a list of available language and its count"""
     cursor.execute(SQL['get-available-languages-and-count'])
-    return [i[0] for i in cursor.fetchall()]
+    return [i[0:2] for i in cursor.fetchall()]
```
hackthehackerman: @karenc Thanks for pointing it out. I think this is fixed now.

cnxarchive/views.py
```diff
@@ -399,6 +399,12 @@ def _get_subject_list(cursor):
         yield subject
 
 
+def _get_available_languages_and_count(cursor):
+    """Return a list of available language and its count"""
+    cursor.execute(SQL['get-available-languages-and-count'])
+    return [i[0:2] for i in cursor.fetchall()]
```
karenc: You can probably just `return cursor.fetchall()` :)

karenc:
When you're done with all the changes, if you could clean up the commits, that'll be great.

You can either use `git rebase -i master` if you're familiar with it or:

```
git reset master
git add cnxarchive/sql/get-available-languages-and-count.sql
git commit -a
git push -f origin HEAD
```

cnxarchive/sql/get-available-languages.sql
```diff
@@ -0,0 +1,11 @@
+-- ###
```
karenc: Are you working on some more changes that use this file?

hackthehackerman: @karenc no, I must have forgotten to delete it. Will do.

hackthehackerman:
@karenc I can rebase and merge my commits into one. But since my commits were already pushed to this branch, I can either push --force to rewrite the history(problematic for others who has pulled this branch before) or I can delete/make a new branch with the compact commit. Which way would you like?

karenc:
Force push is ok for pull requests I think.

hackthehackerman:
@karenc it's ok as in I can do that. However, this pull request is to merge my branch, which lives inside the current repo, into the master branch. If anyone else has pulled the current branch then it might create problems for them later. Since nobody else is working on this branch, I guess it's ok to do that.

karenc:
There are ways to deal with rebased branches, I wouldn't worry too much about it :p

hackthehackerman:
@karenc ok, thanks!

karenc:
Looks great, thank you!

reedstrm:
looks good. Merging.

PR #441, opened 2016-05-20
Switch from auto generated sitemap url to a config setting

hackthehackerman:


cnxarchive/tests/testing.ini
```diff
@@ -4,7 +4,7 @@ db-connection-string = dbname=cnxarchive-testing user=cnxarchive password=cnxarc
 memcache-servers = localhost
 search-cache-expiration = 15
 search-long-cache-expiration = 30
-
+sitemap-destination = http://cnx.org/sitemap.xml
```
pumazi: This looks like a required setting that isn't going to fail at startup. So let's make one minor change. Please add this setting to the `manditory_settings` list in https://github.com/Connexions/cnx-archive/blob/f4522c5087c7a44d6e8726dc91a2adf46ac5a5c9/cnxarchive/__init__.py#L81

karenc: Is it possible to generate the sitemap url?

hackthehackerman: @karenc it's possible but I don't see the benefit of doing that. We want to point the sitemap to another url's sitemap. If we auto generate it by using the root url like: archive.cnx.org ==> cnx.org, then it will break if archive.cnx ever move to another domain.

karenc: Ah right, we need the webview url, not the archive url.

PR #442, opened 2016-05-30
Correctly update submiter/log on collection replublish

hackthehackerman:
The database hook that prepare new collection data when new modules is
inserted did not include the submitter/submitlog data. This commit
added them into the hook so the new collection row will have the
correct data on each minor update.

### DIFF
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -372,8 +372,8 @@ def rebuild_collection_tree(old_collection_ident, new_document_id_map, plpy):
     build_tree(root_node, None)
 
 
-def republish_collection(next_minor_version, collection_ident, plpy,
-                         revised=None):
+def republish_collection(submitter, submitlog, next_minor_version,
+                         collection_ident, plpy, revised=None):
     """Insert a new row for collection_ident with a new version.
 
     Returns the module_ident of the row inserted.
@@ -385,7 +385,7 @@ def republish_collection(next_minor_version, collection_ident, plpy,
         parentauthors, google_analytics, buylink, print_style,
         major_version, minor_version)
       SELECT m.portal_type, m.moduleid, m.uuid, m.version, m.name, m.created,
-        {}, m.abstractid, m.licenseid, m.doctype, m.submitter, m.submitlog,
+        {}, m.abstractid, m.licenseid, m.doctype, $3, $4,
         m.stateid, m.parent, m.language, m.authors, m.maintainers, m.licensors,
         m.parentauthors, m.google_analytics, m.buylink, m.print_style,
         m.major_version, $1
@@ -395,12 +395,13 @@ def republish_collection(next_minor_version, collection_ident, plpy,
     '''
     if revised is None:
         sql = sql.format('CURRENT_TIMESTAMP')
-        types = ('integer', 'integer')
-        params = (next_minor_version, collection_ident)
+        types = ('integer', 'integer', 'text', 'text')
+        params = (next_minor_version, collection_ident, submitter, submitlog)
     else:
-        sql = sql.format('$3')
-        types = ('integer', 'integer', 'timestamp')
-        params = (next_minor_version, collection_ident, revised)
+        sql = sql.format('$5')
+        types = ('integer', 'integer', 'text', 'text', 'timestamp')
+        params = (next_minor_version, collection_ident, submitter, submitlog,
+                  evised)
     plan = plpy.prepare(sql, types)
     new_ident = plpy.execute(plan, params, 1)[0]['module_ident']
 
@@ -464,6 +465,8 @@ def republish_module(td, plpy):
     modified = 'OK'
     moduleid = td['new']['moduleid']
     legacy_version = td['new']['version']
+    submitter = td['new']['submitter']
+    submitlog = td['new']['submitlog']
 
     modified = set_version(portal_type, legacy_version, td)
 
@@ -484,7 +487,8 @@ def republish_module(td, plpy):
     # Module is republished
     for collection_id in get_collections(current_module_ident, plpy):
         minor = next_version(collection_id, plpy)
-        new_ident = republish_collection(minor, collection_id, plpy)
+        new_ident = republish_collection(submitter, submitlog, minor,
+                                         collection_id, plpy)
         rebuild_collection_tree(collection_id, {
             collection_id: new_ident,
             current_module_ident: td['new']['module_ident'],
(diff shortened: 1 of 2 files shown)


karenc:
What does this fix?

reedstrm:
When a module publishes, and triggers containing-collections to rev (minor rev), the commit details (submitter and submitlog) where copied from the last major rev of the collection, rather than from the actual new publish event, that of the module. THis lead to all minor revs having the same commit info. This copies the triggering events info, so the minor rev info on the collection reflects the actual change.

karenc:
Oh I see, in that case, I'd suggest modifying this test to show the problem and that it's fixed: 

https://github.com/Connexions/cnx-archive/blob/15f41bbcc3da0a6ad3f972ffc8bc6a84445423b7/cnxarchive/tests/test_database.py#L984-L1083

hackthehackerman:
The commit on test_database.py is already testing that submitter/submitlog correctly updates during collection replublish.

karenc:
I think the changes in test_database.py is showing that if you give a submitter and submitlog, it will be inserted into the table for the next version.  It doesn't show the bug you're trying to fix.

hackthehackerman:
@karenc Please re-read the test_republish_collection() test case and my changes to it.

cnxarchive/tests/test_database.py
```diff
@@ -780,8 +780,12 @@ def test_republish_collection(self, cursor):
         collection_ident = cursor.fetchone()[0]
         cursor.connection.commit()
 
-        new_ident = republish_collection(
-            3, collection_ident, testing.fake_plpy)
+        replublished_submittor = "replublished_submittor"
+        replublished_submitlog = "replublished_submitlog"
+
+        new_ident = republish_collection(replublished_submittor,
```
karenc: This change to the test shows that if you pass in the submitter and the submitlog, it will end up in the modules table.

It does not show the bug that when a module is published, and a collection is republished, that the collection is going to have the module submitter and submitlog instead of the collection submitter and submitlog.  Do you disagree?

Also, would you like to fix the typos?

hackthehackerman: @karenc Thanks for pointing it out. I see that I cam not checking against the row with the module ident. I will fix it later.

karenc: I'm not sure if I explained it clearly enough, so I'll try again.  The test that you modified calls `republish_collection`, but actually we need to test the thing (the trigger) that calls `republish_collection`, which is https://github.com/Connexions/cnx-archive/blob/15f41bbcc3da0a6ad3f972ffc8bc6a84445423b7/cnxarchive/sql/schema/triggers.sql#L150-L152

That's why I'm suggesting that you also change the test here https://github.com/Connexions/cnx-archive/blob/15f41bbcc3da0a6ad3f972ffc8bc6a84445423b7/cnxarchive/tests/test_database.py#L984-L1083 which is actually testing the trigger.

hackthehackerman: Thank you for your detailed explanation. I will fix it accordingly.

hackthehackerman: @karenc I do think that I am not correctly testing. However, I don't see how it is wrong to test the result by directly calling republish_collection. The bug is not whether the trigger is fired, but whether the trigger is correctly doing what it is supposed to do. Therefore, I am going to modify the test that I originally modified.

hackthehackerman: @karenc Actually, I read my test code again and I think I am doing it correctly. The previous republish_collectioon method ignores submitter/submitlog. After my change, it takes the submitter/submitlog as arguments and insert them into the correct row. In the test I modified, I checked if the submitter/submitlog is correctly inserted into the row. Correctly inserted data solves the bug mentioned in [this ticket](https://trello.com/c/aXesAJtk/13-improve-full-archive-search-w-in-book-search-technology). So I think my commit is not doing anything wrong besides the typo.

karenc: I'm not saying that you did anything wrong.  You added some arguments to `republish_collection`, you should make sure those changes work, which you did.

cnxarchive/database.py
```diff
@@ -484,7 +487,8 @@ def republish_module(td, plpy):
     # Module is republished
     for collection_id in get_collections(current_module_ident, plpy):
         minor = next_version(collection_id, plpy)
-        new_ident = republish_collection(minor, collection_id, plpy)
+        new_ident = republish_collection(submitter, submitlog, minor,
```
karenc: This is the part that is not covered by the tests you modified.  I wanted to make sure that the correct `submitter` and `submitlog` are being used in the collection minor version.

karenc:
If we have a commit like this before your commit:

```diff
commit 551a70b01de4a5b3dca661ec2f198f7311051c42
Author: karen chan <karen@karen-chan.com>
Date:   Tue Jun 7 20:28:02 2016 +0100

    Add asserts for submitter & submitlog for collection republication

diff --git a/cnxarchive/tests/test_database.py b/cnxarchive/tests/test_database.py
index f883a0b..4d0ec29 100644
--- a/cnxarchive/tests/test_database.py
+++ b/cnxarchive/tests/test_database.py
@@ -991,7 +991,7 @@ ALTER TABLE modules DISABLE TRIGGER module_published""")
          language, parent)
         VALUES ('m42955', 'Module', '1.2', 'Preface to College Physics',
         '2013-09-13 15:10:43.000000+02' , '2013-09-13 15:10:43.000000+02',
-        NULL, NULL, NULL, 1, NULL, 11, '', NULL, '',
+        NULL, NULL, NULL, 1, NULL, 11, '', 'karenc', 'I changed something',
         'en', NULL) RETURNING module_ident''')
         new_module_ident = cursor.fetchone()[0]
 
@@ -1017,6 +1017,8 @@ ALTER TABLE modules DISABLE TRIGGER module_published""")
 
         self.assertEqual(results[1], 'Collection')  # portal_type
         self.assertEqual(results[5], 'Derived Copy of College Physics')  # name
+        self.assertEqual(results[11], 'OpenStaxCollege')  # submitter
+        self.assertEqual(results[12], 'New version 1.7')  # submitlog
         self.assertEqual(results[-3], 1)  # major_version
         self.assertEqual(results[-2], 2)  # minor_version
         self.assertEqual(results[-1], None)  # print_style
```

We'll be able to show that after your change, the collection new minor version uses the module's submitter (`karenc`) and submitlog (`I changed something`) instead of the collection one (`OpenStaxCollege` and `New version 1.7`).

reedstrm:
@hackthehackerman could you work in a commit like @karenc asked for, above? THen we can get this merged, and out w/ the next release.

hackthehackerman:
@reedstrm Will finished by end of Friday.

hackthehackerman:
Closing because https://github.com/Connexions/cnx-archive/pull/444

PR #443, opened 2016-06-09
Use cnx-db init instead of cnx-archive-initdb

karenc:
Depends on pumazi/cnx-db#2 and pumazi/cnx-db#3.

~~**Important**: Please remove the last commit before merging.~~


pumazi:
The tests have been rerun with the latest release of cnx-db `==0.2.6`, which includes the sub-collection uuid changes.

PR #445, opened 2016-06-24
Post publication trigger

karenc:
Also see https://github.com/openstax/napkin-notes/blob/master/pumazi/cnx/design--post-publication-states.md


cnxarchive/sql/migrations/20160622095217_add_modules_stateid_default.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+
+
+def up(cursor):
+    cursor.execute("""\
+ALTER TABLE modules
+  ALTER COLUMN stateid SET DEFAULT 1""")
+
```
reedstrm: Hmm, this might take an exclusive lock on the table to set the default. I'll have to check.

cnxarchive/sql/schema/triggers.sql
```diff
@@ -1,9 +1,10 @@
 CREATE OR REPLACE FUNCTION update_latest() RETURNS trigger AS '
 BEGIN
-  IF TG_OP = ''INSERT'' AND
+  IF (TG_OP = ''INSERT'' OR TG_OP = ''UPDATE'') AND
           NEW.revised >= ((SELECT revised FROM modules
               WHERE uuid = NEW.uuid ORDER BY revised DESC LIMIT 1)
-              UNION ALL VALUES (NEW.revised) LIMIT 1) THEN
+              UNION ALL VALUES (NEW.revised) LIMIT 1) AND
+          NEW.stateid = 1 THEN
```
reedstrm: so, published content will not appear until it has been processed, at all. Hmm, I think that's probably correct.

reedstrm:
This looks good. It's intended to only roll out with the companion listener/processing code, right?

karenc:
@reedstrm Yes, I just created a WIP PR in cnx-publishing.  The changes in the `update_latest` trigger caused some test failure that are a bit hard to fix so taking some time...

cnxarchive/sql/migrations/20160622095217_add_modules_stateid_default.py
reedstrm: Yup, I checked, it takes an AccessExclusiveLock, which blocks parallel reads. This might be a bit tricky to rollout on production. Not really any other way to do it, though.

karenc:
Ready for review.

PR #448, opened 2016-07-30
Add cnx-archive-init_venv to set up python venv on postgres

karenc:
The script is copied from
https://github.com/pumazi/cnx-db/blob/454ee6a67ae265ac5dc1010248b9f3f1c7fd1722/cnxdb/init/main.py

Once we use cnx-db in cnx-archive, this script can be removed.

---

This pull request includes commits from #447.


pumazi:
The venv init and re-init is now in `cnx-db==0.2.0`. We may not need this now?

karenc:
I was originally doing this because cte-dev venv was not set up after copying the production database.  I guess we don't need this anymore so I'll close this.

PR #449, opened 2016-08-03
Fix test to correct an error when cnx-archive is only editably installed

pumazi:
This is the difference between when `setup.py develop` and `setup.py install`
are used. One produces an editable install, while the other installs
a copy of the project in site-packages. The test was failing if the package
is development install but not a fixed install.


karenc:
Thanks for fixing it!

PR #450, opened 2016-08-04
Move modulestates to schema initialization

karenc:
Module states should really be part of the database initialization, it's
not data to be added later.  It is necessary for the application, so
moving it to ``cnxarchive/sql/schema/``.


karenc:
See also https://github.com/Connexions/cnx-publishing/pull/136

pumazi:
For the sake of consistency, could you put these in a file within the ``schema/constants`` directory?

karenc:
Oh I didn't realize we have a `schema/constants` directory... :sweat:
I've moved the file.

PR #454, opened 2016-08-24
db migration to add venv support

reedstrm:
After creating a new deploy, and pushing a copy of production data to it, we run into the problem that virtualenv support is only added during initdb when two conditions are met: archive is running inside a venv, and the db connection is local. Production does not meet these requirements, so the data-dump from there does not have the venv support in place. This migration is written to be idempotent, so it will install the venv activation code only if it's not already there. This should allow it to run against datadumps from production as well as from other venv based installs.

### DIFF
--- a/cnxarchive/sql/migrations/20160824022713_add_venv_support.py
+++ b/cnxarchive/sql/migrations/20160824022713_add_venv_support.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+
+
+def up(cursor):
+    """Install venv schema and session_preload if in a virtualenv."""
+    import sys
+
+    if hasattr(sys, 'real_prefix'):
+        activate_path = os.path.join(
+            os.path.realpath(sys.prefix),
+            'bin/activate_this.py')
+        cursor.execute("SELECT current_database();")
+        db_name = cursor.fetchone()[0]
+
+        cursor.execute("""SELECT schema_name FROM
+                          information_schema.schemata
+                          WHERE schema_name = 'venv';""")
+        schema_exists = cursor.fetchall()
+
+        if not schema_exists:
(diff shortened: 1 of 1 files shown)


pumazi:
I believe a similar fix was proposed in #448. It was pushed off because cnx-db includes this functionality. There is no reason we couldn't use cnx-db to achieve the same goal without having it as a fixated dependency of this application.

pumazi:
Yes, putting cnx-db into cnx-deploy would be the correct place.

This project is in my personal repo because it started as an experiment. We've now moved beyond the experimental, so it's ready to be moved.

We've pulled back on including cnx-db in cnx-archive and cnx-publishing because it would add to the already monstrous release. This package is only in it's infancy, but would move the schema for cnx-archive and cnx-publishing. While this isn't a huge risk, it does change the the project significantly. Thus, we pulled it back with the assumption that it would easily fit into the next release. However, since both #448 and this pull request attempt to make the same change, it must be really really needed. So, we'll need to work towards getting the cnx-db stuff integrated. (see also #443 and Connexions/cnx-publishing#129) Or we accept this pull request or #448.

reedstrm:
perhaps the right answer for 'right now' is merge this PR (which leverages the migration to put the venv support in place, no extra step needed) marked to remove it when the cnx-db-ification change lands.

karenc:
It would be nice if we can use cnx-db in cnx-deploy now...

reedstrm:
if cnx-db venv correct the problem I had (no venv after push_data) I see no blocker from putting it in just for that, even without replacing the initdb calls.

karenc:
I hope this is going to fix the venv problem: https://github.com/Connexions/cnx-deploy/pull/83

karenc:
Connexions/cnx-deploy#83 is merged, venv will be initialized every time the main playbook is run.

PR #455, opened 2016-09-02
Collection minor version files

reedstrm:
We now attach certain files to collections. Preserve them.


reedstrm:
fixes #456

karenc:
Thanks for fixing this :smile:

PR #457, opened 2016-09-14
In collated book search

reedstrm:
Adds the fulltext index infrastructure needed to make collated book searches work.

### DIFF
--- a/cnxarchive/tests/data/README.rst
+++ b/cnxarchive/tests/data/README.rst
@@ -1,20 +1,32 @@
 :data.sql:
     Generated via::
         cnx-archive-initdb cnxarchive/tests/testing.ini --with-example-data
-        # Initialize the schema manager
-        dbmigrator --config cnxarchive/tests/testing.ini init
-        # Create the schema migration step
-        dbmigrator generate <description of the change>
-        # Make needed changes via db-migrator migrations
-        dbmigrator --config cnxarchive/tests/testing.ini migrate
-        pg_dump -a -T licenses -T tags -T roles --disable-triggers --inserts cnxarchive-testing > data.sql
+        pg_dump -a -T schema_migrations -T licenses -T tags -T roles -T modulestates -T service_states --disable-triggers --inserts cnxarchive-testing > cnxarchive/tests/data/data.sql
 
-
-
-:legacy-data.sql: Contains a data set extracted from a legacy database
-    structure that was populated using cnx-populate. This
-    data is a SQL dump using::
-
-        pg_dump --data-only -T licenses -T tags -T trees \
-          --inserts --disable-triggers \
-          -d $DB_NAME > legacy-data.sql
+        If you have a a feature branch with schema and/or data changes, you'll
+        need to build a migration. Which is basically a schema-diff. The workflow below
+        uses tools from Pyrseas https://github.com/perseas/Pyrseas. It's slightly simplier if
+        you can make the migrations steps by hand.
+            git checkout myfeature
+            cnx-archive-initdb cnxarchive/tests/testing.ini
+            # dump a schema
+            dbtoyaml cnxarchive-testing -o cnxarchive-myfeature.yml
+            dropdb cnxarchive-testing; createdb -O cnxarchive cnxarchive-testing
+            git checkout master
+            cnx-archive-initdb cnxarchive/tests/testing.ini
+            # dump a schema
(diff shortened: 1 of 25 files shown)


cnxarchive/tests/data/README.rst
```diff
@@ -1,20 +1,28 @@
...
+        need to build a migration. Which is basically a schema-diff.
+            git checkout myfeature
+            cnx-archive-initdb cnxarchive/tests/testing.ini
+            # dump a schema
+            dbtoyaml cnxarchive-testing -o cnxarchive-myfeature.yml
+            dropdb cnxarchive-testing; createdb -O cnxarchive cnxarchive-testing
+            git checkout master
+            cnx-archive-initdb cnxarchive/tests/testing.ini
+            # dump a schema
+            dbtoyaml cnxarchive-testing -o cnxarchive-master.yml
```
pumazi: You'll want to mention where these tools come from. (I ended up searching for 'pysears dbtoyaml' and 'pysiers dbtoyaml' and finally come across the actual name.) Though, the use of pyrseas seems like overkill here.

karenc: Perhaps you can put the auto generation of migrations somewhere else (in another section?) and just leave this for the simpler case of hand written migrations. ;D

pumazi:
Doing a separate copy commit for `*_in_collated_book_search` sql files would have been good. This would have enabled the reviewer and historian to see the specific differences between the files. Just something to keep in mind for the future.

cnxarchive/tests/data/README.rst
reedstrm: yeah, thought about that. The existing steps where unclear, so I documented what I did for this case. :-)

cnxarchive/views.py
```diff
@@ -663,11 +663,14 @@ def in_book_search(request):
 
     settings = get_current_registry().settings
     connection_string = settings[config.CONNECTION_STRING]
-    statement = SQL['get-in-book-search']
+    statement = SQL['get-in-collated-book-search']
     with psycopg2.connect(connection_string) as db_connection:
         with db_connection.cursor() as cursor:
             cursor.execute(statement, args)
             res = cursor.fetchall()
+            if not res:  # Perhaps it's not collated
```
reedstrm: TBH, I'm not 100% comfy with this idiom here, since there may be other reasons that it's not returning a result. Perhaps I should have gone with the 'do an extra  round trip to the DB to test for is_collated, then call the right one' here.

cnxarchive/views.py
```diff
@@ -719,11 +722,15 @@ def in_book_search_highlighted_results(request):
 
     settings = get_current_registry().settings
     connection_string = settings[config.CONNECTION_STRING]
-    statement = SQL['get-in-book-search-full-page']
+    statement = SQL['get-in-collated-book-search-full-page']
```
reedstrm: and here

cnxarchive/tests/data/README.rst
reedstrm: added note about and link to Pyrseas. Tested steps by updating the migration for the new stuff I just added. :-)

reedstrm:
I renamed the function, and also put it into every SQL script that might benefit from using the new index.

reedstrm:
BTW, tree_to_json is now approx 10x faster (typically 10-20ms instead of 120-320ms)

pumazi:
It's clear this all works and a vast improvement has been made, because the tests pass and the speed has improved, but the changes are really intermingled. For example, the `module_version` function should be it's own separate commit that could precede the feature commits. This would make it clearer to analyze what's happening during each changeset. Another example, there is a `README.md` change buried in commit about updating the migration, which is unrelated. I don't want this to sound critical; and hopefully it doesn't... I have the feeling that much of this could have been condensed to 3-4 very specific commits, each of which could have been analyzed separate from the others. There are a lot of changes here (in 26 different files), so having the logical concerns separate would have been extremely useful. (Also, looking into the future, I really wouldn't want to piece together what was happening in these changes and hope that I don't have to. 😕 )

karenc:
I'll attempt to re-organize the commits.

reedstrm:
Yes, the feature could have been better organized. In my defence, I've been chastized for PRs that don't squash my work history, in the past. :-) I just force-pushed a reorg, hope you like it better

reedstrm:
The first commit is still a bit large, but I really don't want to detangle it more. I expanded all the commit messages to explain what's going on a bit.

PR #459, opened 2016-09-21
In collated search 2

karenc:
The changes are exactly the same as #457 just expanded to 11 commits :sweat_smile:


reedstrm:
Wow, @karenc, that's phenomenal refactor of my somewhat scattered changes! It even looks all logical now. :-) So, the idea is to merge this as-is, so this set of more-coherent changes lands on master history? My only complaint is that it _looks_ like a logical dev workflow that never actually happened: I'm much more "bash around in the code until it works" than this implies. :-/

PR #460, opened 2016-09-23
Collated search fixes

reedstrm:
Fixing two problems with collated search. First, the results where wrong - failed to limit the matches to the cooked version of the page that was in that specific book/context - this lead to bogus results as soon as I used it on a server with more than one cooked book using the same modules. Second, the new performance indices where not actually being used due to some typecasting. 
Includes tests, and a db migration

### DIFF
--- a/cnxarchive/sql/migrations/20160923184915_correct_uuid_index.py
+++ b/cnxarchive/sql/migrations/20160923184915_correct_uuid_index.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+
+
+def up(cursor):
+    # Create correct index, drop less useful one
+    cursor.execute("""
+    CREATE INDEX modules_uuid_txt_version_idx ON modules
+    (cast(uuid as text), module_version(major_version, minor_version));
+    CREATE INDEX latest_modules_uuid_txt_version_idx ON latest_modules
+    (cast(uuid as text), module_version(major_version, minor_version));
+    DROP INDEX modules_uuid_version_idx;
+    DROP INDEX latest_modules_uuid_version_idx;
+    """)
+
+
+def down(cursor):
+    # Drop correct index, create less useful one
+    cursor.execute("""
+    CREATE INDEX modules_uuid_version_idx ON modules
+    (uuid, module_version(major_version, minor_version));
+    CREATE INDEX latest_modules_uuid_version_idx ON latest_modules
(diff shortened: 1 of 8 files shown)


reedstrm:
Grr, I guess I should extend the test data to catch this case - add another cooked book that shares a module with the existing one. I really don't have the round-to-its for that right now.

reedstrm:
Adding test content, and more tests, finding more bugs. Applied WIP, so don't merge yet.

reedstrm:
All fixed, all tests in place. Ready for review and merge.

cnxarchive/sql/migrations/20160923184915_correct_uuid_index.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+
+
+def up(cursor):
+    # Create correct index, drop less useful one
+    cursor.execute("""
+    CREATE INDEX modules_uuid_txt_version_idx ON modules
```
pumazi: This is `modules_uuid_tExt_version_idx` in the down step. Typo in one or the other?

reedstrm: yup, up is right (down didn't get much testing, obviously) will fix.

reedstrm: fixed (with --fixup and --autosquash, so history is modified --force-with-lease deployed)

cnxarchive/sql/migrations/20160923184915_correct_uuid_index.py
```diff
@@ -0,0 +1,25 @@
...
+
+def down(cursor):
+    # Drop correct index, create less useful one
+    cursor.execute("""
+    CREATE INDEX modules_uuid_version_idx ON modules
+    (uuid, module_version(major_version, minor_version));
+    CREATE INDEX latest_modules_uuid_version_idx ON latest_modules
+    (uuid, module_version(major_version, minor_version));
+    DROP INDEX modules_uuid_txt_version_idx;
+    DROP INDEX latest_modules_uuid_text_version_idx;
```
pumazi: And this one? 😬

reedstrm: that one too, now.

PR #462, opened 2016-10-25
Create SubCollection metadata objects when shredding collxml

reedstrm:
First part of adding metadata info for subcollections. This creates them when shredding collxml. Need to check other paths that insert into trees (trigger for minor-version creation, cnx-publishing path). Also needs tests (shred_collxml seems to have no tests at all)


reedstrm:
This now depends on epub PR https://github.com/Connexions/cnx-epub/pull/109

pumazi:
Those are some good looking migrations. 👍

pumazi:
I was just looking back on this. @reedstrm, you created a good pull request title on this one. 👍  And it's a pull request that is well scoped to the specific subject area. 💯

PR #464, opened 2016-12-12
encode shortid as well in export epub metadata

reedstrm:
Depends on https://github.com/Connexions/cnx-epub/pull/106


reedstrm:
@phil how about this one, too? :-)

PR #465, opened 2017-01-12
switch from coveralls to codecov

philschatz:
As part of https://github.com/openstax/napkin-notes/issues/68 this switches from coveralls to codecov.

I was not sure about a couple things but took some defaults:

- When should coveralls make comments on the PR and what should be included
- I used this recommendation from codecov.io: https://github.com/codecov/example-python

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
   - cd rhaptos.cnxmlutils && python setup.py install && cd ..
   # Scripts get installed to /usr/local/bin
   - pip install coverage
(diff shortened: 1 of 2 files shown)


pumazi:
#467 resolves the test failures you are experiencing.

pumazi:
@philschatz Can you look at this? There doesn't seem to be any response from codecov after the successful run.

philschatz:
Thanks! Looks like codecov took a little while but eventually reported, right?

PR #466, opened 2017-01-18
Add a dry-run option

DarkPrinceFrost:


pumazi:
We just add a test or two, then this should be good to go.

PR #468, opened 2017-02-03
properly find sql files for migrations

reedstrm:


cnxarchive/sql/migrations/20170201213850_add-subcollection-metadata.py
```diff
@@ -1,4 +1,17 @@
...
+
+from contextlib import contextmanager
+
+
+@contextmanager
+def open_here(filepath, *args, **kwargs):
+    """open a file relative to this files location"""
+
+    here = os.path.abspath(os.path.dirname(__file__))
+    fp = open(os.path.join(here, filepath), *args, **kwargs)
```
pumazi: I'm pretty sure `open` returns a context manager. So it's probably not necessary to wrap this with another context manager, but this works all the same.

PR #472, opened 2017-02-14
Correctly assign subcollection (chapter) ids

reedstrm:
Because of the way published content was being parsed into a model,
the ids of a baked chapter's content was being set to the same
as the first page within the chapter. This fixes that.


reedstrm:
While updating expected test output, I've decided that this fix, while it does work, is in th wrong place. I'll dig deeper into where the metadata on the binder is being processed, to avoid adding cnx-archive-* to the actual binder objects - those names should only be in the HTML, AFAIAC.

reedstrm:
Changed my mind: these are valid metadata attributes. However, when presented as cnx-archive-*, they should be fetchable from cnx-archive. The transparent binder id of 'subcol' is not, so exclude it.

PR #473, opened 2017-02-27
Export baked epub

reedstrm:
Need this for  actually getting the bits out of the server to deliver to outside sources.

### DIFF
--- a/cnxarchive/scripts/export_epub/modeling.py
+++ b/cnxarchive/scripts/export_epub/modeling.py
@@ -25,8 +25,11 @@ from cnxarchive.scripts.export_epub.db import (
     )
 
 
-def document_factory(ident_hash, context=None):
+def document_factory(ident_hash, context=None, baked=False):
     metadata = get_metadata(ident_hash)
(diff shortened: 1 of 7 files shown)


cnxarchive/scripts/export_epub/modeling.py
```diff
@@ -25,8 +25,11 @@
     )
 
 
-def document_factory(ident_hash, context=None):
+def document_factory(ident_hash, context=None, baked=False):
```
mwharrison: I think you need `baked=None` or your check below should be `if baked and context is False`, otherwise it will always be False and the RuntimeError will never be raised.

reedstrm: Nope, `is` binds tighter than `and`, so `True and None is None` is `True`.

reedstrm:
@mwharrison how's it look now? (Yay the tests pass!)

reedstrm:
I did some fixup! commits, so i think a rebase-n-merge will squash those.

PR #474, opened 2017-03-17
Update README to mention webview and Python version

kerwinso:


README.rst
```diff
@@ -176,7 +176,8 @@ command::
 
 To run the application, use the ``paste`` script with the ``serve`` command.
 (The paste script and serve command come from ``PasteScript`` and
-``PasteDeploy``, respectively.)
+``PasteDeploy``, respectively, which will need to be installed: ``pip install pastescript``
+``pip install pastedeploy``.)
```
pumazi: It's showing the age of this documentation, but this is no longer true. The real command to use is Pyramid's `pserve`, which deep down contains forked bits of PasteScript and PasteDeploy. So the statement above isn't altogether false, but clarification is definitely needed.

README.rst
```diff
@@ -2,9 +2,9 @@ Connexions Archive Repository
 =============================
 
 This is an archive for Connexions documents. It holds published
-documents and collections of documents. It is accessible to the public via
-a read-only API. It has an optional write API for publishing content
-from an unpublished repository.
+documents and collections of documents and is designed to work together with `webview <https://github.com/Connexions/webview>`_.
+It is accessible to the public via a read-only API. It has an optional write API for publishing content
+from an unpublished repository. It runs best in Python 2.7.
```
pumazi: It only runs on python 2.7 at this time.

kerwinso: ok

README.rst
kerwinso: I don't know much about `pserve`, so feel free to suggest a revision that is clearer. If it makes sense to reject this particular change, that's fine.

reedstrm:
The changes in this PR are now correct. Fixing the `paste serve` to be `pserve` is another issue.

PR #475, opened 2017-04-05
Shortids in tree

reedstrm:
Minimal fix to get the cnx-archive-* attribs into the nav tree of the epub HTML file, for TEG/TEA. I think a complete overhaul of how I bodged shortid into epub is needed, actually, but out-of-scope for right now.

Aside: My thoughts on that are to remove shortId knowledge completely from cnx-epub, make sure id is always == uuid, and version lives in version (so hash == id@version), then deal w/ computing/adding shortId as a display property in cnx-archive-* tools themselves. May need to consider some tricks for having cnx-archive pass in info to cnx-epub about what metadata to copy out to nodes (or not - could be it just needs to carry metadata for every node, and pass it back to archive for processing, though would require passing rendering templates into cnx-epub from archive, to be clean,  I think.)


reedstrm:
squash merge, please

PR #476, opened 2017-04-06
rev subcol minor id as well

reedstrm:
This properly bumps the minor version numbers of subcollections impacted when a module is published, as well as the collection(s) it is a part of.

### DIFF
--- a/cnxarchive/tests/test_database.py
+++ b/cnxarchive/tests/test_database.py
@@ -1028,11 +1028,15 @@ ALTER TABLE modules DISABLE TRIGGER module_published""")
 
     @testing.db_connect
     def test_module(self, cursor):
-        # Create a fake collated tree for College Physics which contains the
-        # module that is going to have a new version
+        # Create a fake collated tree for College Physics, with a chapter,
+        # which contains the module that is going to have a new version
         cursor.execute("""\
 INSERT INTO trees (parent_id, documentid, is_collated)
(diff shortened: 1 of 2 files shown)


reedstrm:
needs a tweak to test data to get more coverage, otherwise is complete.

reedstrm:
@pumazi This patch changes the trigger code, which codecoverage doesn't track, since it runs in the trigger process, not this process. I just added tests that definitely run all the new subcol manipulating code in the trigger, since it `Asserts` their values. Please merge without codecov passing.

reedstrm:
rebased

cnxarchive/tests/test_database.py
```diff
@@ -1081,8 +1086,41 @@ def test_module(self, cursor):
         self.assertEqual(results[1], 'Collection')  # portal_type
         self.assertEqual(results[5], '<span style="color:red;">Derived</span>'
                                      ' Copy of College <i>Physics</i>')  # name
-        self.assertEqual(results[11], 'karenc')  # submitter
-        self.assertEqual(results[12], 'I changed something')  # submitlog
+        self.assertEqual(results[11], 'reedstrm')  # submitter
```
pumazi: 😛

reedstrm: yeah, I almost left, it, but I changed which module was being published, so …

PR #482, opened 2017-05-06
Subcol minor version additional tests

reedstrm:
This restores the codecov lost in the  in #476


cnxarchive/tests/testing.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
pumazi: Why did you remove the encoding?

PR #485, opened 2017-05-30
Fix to disallow special char and < 3 char searches

pumazi:
It's not exactly pretty, but it should work. The search view is a mess, so adding this condition to that mess isn't going to make it that much worse. 😛

### DIFF
--- a/cnxarchive/search.py
+++ b/cnxarchive/search.py
@@ -37,6 +37,7 @@ with open(os.path.join(here, 'data', 'common-english-words.txt'), 'r') as f:
                  [chr(i) for i in range(ord('a'), ord('z') + 1)])
 WILDCARD_KEYWORD = 'text'
 VALID_FILTER_KEYWORDS = ('type', 'pubYear', 'authorID', 'submitterID')
+SEARCH_SPECIAL_CHARS = '"\',:'
 # The maximum number of keywords and authors to return in the search result
 # counts
 MAX_VALUES_FOR_KEYWORDS = 100
--- a/cnxarchive/tests/test_views.py
+++ b/cnxarchive/tests/test_views.py
@@ -2103,8 +2103,63 @@ INSERT INTO trees (nodeid, parent_id, title, childorder, is_collated)
                 },
             }))
 
+    def test_search_with_only_special_chars(self):
+        # Disallow searches that contain only special chars ["',:]
+        from ..search import SEARCH_SPECIAL_CHARS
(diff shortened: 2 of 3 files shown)


pumazi:
Since this PR is a fork from `v2.5.1`, we'll need to ignore some issues that have been corrected since then. The failing tests (at least on the push) are due to a change in the encoding procedure in webob a dependency of pyramid.

cnxarchive/tests/test_views.py
```diff
@@ -2103,8 +2103,63 @@ def test_search_whitespace(self):
...
+            u'results': {
+                u'items': [],
+                u'total': 0,
+                u'limits': [],
+                },
+            }))
+
     def test_search_utf8(self):
-        self.request.params = {'q': '"你好"'}
+        self.request.params = {'q': '"今天你好吗"'}
```
reedstrm: Hmm, I bet the 2-char Chinese search is sufficiently meaningful that we shouldn't disallow it. I'd bet single-char are even useful.

cnxarchive/search.py
```diff
@@ -37,6 +37,7 @@
                  [chr(i) for i in range(ord('a'), ord('z') + 1)])
 WILDCARD_KEYWORD = 'text'
 VALID_FILTER_KEYWORDS = ('type', 'pubYear', 'authorID', 'submitterID')
+SEARCH_SPECIAL_CHARS = '"\',:'
```
reedstrm: Hmm, currently we can search for contractions like "don't" which can actually be useful.

pumazi: This won't affect how searches are done. Stripped, `don't` will be `dont` which is a valid search.

cnxarchive/tests/test_views.py
pumazi: I get what you are saying here, but I don't think there is a simple fix.

reedstrm:
Perhaps make this a disallow the completely blank query, only?

cnxarchive/search.py
reedstrm: Ah right, strip() only removes from the extremes of the string. so "" and '' and :: and ,, would all string to nothing. How about we keep that, but change the numeric limit to < 2, or remove it completely? I'm particularly concerned about single-char Chinese being useful searches. Two char are definitely useful

reedstrm: Should we add `space` to the SEARCH_SPECIAL_CHARS, so that `" "` collapses as  well?

PR #488, opened 2017-06-05
Recent rss

sisiz:
creating an rss feed for cnx-archive that pulls the most recent modules/Collections from the latest_modules table in the database.
implements query to set number of modules to see at once (default 10), start module for batching (default 0 which is most recent module), and type (by default modules and collections).


cnxarchive/views.py
```diff
@@ -981,3 +1016,47 @@ def robots(request):
     resp.expires = html_date(exp_time)
     resp.body = robots_dot_txt.to_string()
     return resp
+
+@view_config(route_name='recent', request_method='GET',
+             accept="application/json",
+             renderer='templates/recent.rss') #renderer: cnxarchive.views:templates/recent.html?? also what should the permission actuall be??
```
pumazi: Let's remove the comment here. Also, let's remove the `accept` argument, it's not useful in this case.

cnxarchive/views.py
```diff
@@ -981,3 +1016,47 @@ def robots(request):
...
+    for i in range(len(search_results)):
+        latest_modules.append(dict(zip(titles, search_results[i])))
+    for module in latest_modules:
+        # module['revised'] = format_date(module['revised'])
+        module['revised'] = html_date(module['revised'])
+        module['authors'] = format_author(module['authors'], settings)
+        module['abstract'] = module['abstract'].decode('utf-8')
+        module['name'] = module['name'].decode('utf-8')
+
+    return {"latest_modules": latest_modules}
```
pumazi: How about we also set the response `Content-type` header? This can be done via `request.response.content_type = 'application/rss+xml'`.

cnxarchive/views.py
```diff
@@ -981,3 +1016,47 @@ def robots(request):
     resp.expires = html_date(exp_time)
     resp.body = robots_dot_txt.to_string()
     return resp
+
+@view_config(route_name='recent', request_method='GET',
+             accept="application/json",
+             renderer='templates/recent.rss') #renderer: cnxarchive.views:templates/recent.html?? also what should the permission actuall be??
+def recent(request):
+    # setting the query variables
+    query_vars = get_query_vars(request.query_string)
```
pumazi: It's not necessary to parse this yourself, though good job in doing so. See the [pyramid.request.Request.GET documentation](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.GET)

An example usage of this might be `num_entries = request.GET.get('number', 10)`

cnxarchive/tests/test_views.py
```diff
@@ -3051,3 +3051,18 @@ def patched_now_side_effect(timezone):
...
+        from ..views import recent
+        recent = recent(self.request)
+        # # note: recent is the dict that is returned by recent function
+
+        self.assertEqual(len(recent['latest_modules']), 10)
+        for module in recent['latest_modules']:
+            keys = module.keys()
+            keys.sort()
+            self.assertEqual(keys,
+                             ["abstract", "authors", "link", "name", "revised"])
```
pumazi: This is a good start. 👍  You may also want to test that you are actually getting the correctly ordered results.

cnxarchive/views.py
```diff
@@ -428,6 +428,41 @@ def _get_licenses(cursor):
...
+    full names and returns a list of the full names as a string.
+    """
+    personids_string = "("
+    for i in range(len(personids) - 1):
+        personids_string += "'{},'".format(peronids[i])
+    personids_string += "'{}')".format(personids[-1])
+    statement = """
+                SELECT fullname
+                FROM persons
+                WHERE personid IN {};
```
pumazi: Would using an `= ANY <ARRAY[]>` here work? It might condense the code, because you wouldn't need to preformat the list, instead psycopg2 will translate the `list` type to a native sql `ARRAY` type.

cnxarchive/views.py
```diff
@@ -428,6 +428,41 @@ def _get_licenses(cursor):
...
+                """.format(personids_string)
+    with psycopg2.connect(settings[config.CONNECTION_STRING]) as db_connection:
+            with db_connection.cursor() as cursor:
+                cursor.execute(statement)
+                authors_list = cursor.fetchall()
+    formated_string = ""
+    for i in range(len(authors_list[0]) - 1):
+        formated_string += (authors_list[0][i]) + ","
```
pumazi: Would a `', '.join(authors_list)` do the same thing that you've done here?

sisiz: yeah, fixing that now

cnxarchive/views.py
sisiz: Is this what you meant? because psycopg2 isn't translating this, or is there another way to get it to translate correctly     
`statement = """
                SELECT fullname
                FROM persons
                WHERE personid = ANY {};
                """.format(personids)
`

cnxarchive/views.py
sisiz: just add that line in?

cnxarchive/tests/test_views.py
sisiz: Added check to make sure the modules were returned in the correct order by date

cnxarchive/views.py
```diff
@@ -981,3 +1016,47 @@ def robots(request):
...
+    statement = """
+                SELECT name, revised, authors, abstract,
+                'http://cnx.org/contents/'||ident_hash( uuid, major_version, minor_version) AS link
+                FROM latest_modules
+                JOIN abstracts ON latest_modules.abstractid = abstracts.abstractid
+                WHERE portal_type in {}
+                ORDER BY revised DESC
+                LIMIT {} OFFSET {};
+                """.format(portal_type, num_entries, start_entry)
+    with psycopg2.connect(settings[config.CONNECTION_STRING]) as db_connection:
```
reedstrm: Really important to do the param passing in execute, here. `number`, `start` and `type` are all passed in from the URL, so contain potentially hostile values. Needs all the `{}` replaced with `%s` then call:
`cursor.execute(statement, vars=portal_type, num_entries, start_entry))`

cnxarchive/views.py
```diff
@@ -981,3 +1016,47 @@ def robots(request):
...
+
+@view_config(route_name='recent', request_method='GET',
+             accept="application/json",
+             renderer='templates/recent.rss') #renderer: cnxarchive.views:templates/recent.html?? also what should the permission actuall be??
+def recent(request):
+    # setting the query variables
+    query_vars = get_query_vars(request.query_string)
+    num_entries = 10
+    start_entry = 0
+    portal_type = "('Collection', 'Module')"
```
reedstrm: make this a python array, and convert the 'portal_type in ' to 'portal_type = ANY'

cnxarchive/views.py
```diff
@@ -428,6 +428,41 @@ def _get_licenses(cursor):
...
+
+
+def format_author(personids, settings):
+    """
+    Takes a list of personid's and searches in the persons table to get their
+    full names and returns a list of the full names as a string.
+    """
+    personids_string = "("
+    for i in range(len(personids) - 1):
+        personids_string += "'{},'".format(peronids[i])
```
reedstrm: Also, the spacer there is `"'{},'"` it should be `"'{}',"`. Otherwise you end up with `['id1,''id2']` rather than `['id1','id2']`

reedstrm:
http://initd.org/psycopg/docs/usage.html

cnxarchive/views.py
pumazi: Yep. Just add that line somewhere before the function return.

sisiz:
Added in the psycopg2 parameter passing.
fixed the portal type parameter to work, but the number and start were working for me before, so not sure what exactly is wrong with them and if it's still broken

cnxarchive/templates/recent.rss
```diff
@@ -0,0 +1,37 @@
...
+            &lt;/div&gt;
+          &lt;/h3&gt;
+          &lt;div class="feedEntryContent"&gt;
+            &lt;b&gt;
+                  {{module.authors}}
+            &lt;/b&gt;
+            &lt;br&gt;&lt;/br&gt;
+               {{module.abstract}}
+            &lt;br&gt;&lt;/br&gt;
+            &lt;img src="http://i.creativecommons.org/l/by/4.0/80x15.png" />
```
pumazi: Should the `/>` be encoded as well?

cnxarchive/views.py
```diff
@@ -981,3 +1008,46 @@ def robots(request):
     resp.expires = html_date(exp_time)
     resp.body = robots_dot_txt.to_string()
     return resp
+
+
+@view_config(route_name='recent', request_method='GET',
+             accept="application/json",
```
pumazi: This `accept` should be removed.

cnxarchive/views.py
```diff
@@ -981,3 +1008,46 @@ def robots(request):
...
+    start_entry = request.GET.get('start', 0)
+    portal_type = request.GET.get('type', ['Collection', 'Module'])
+    if portal_type != ['Collection', 'Module']:
+        portal_type = [portal_type]
+    # search the database
+    settings = request.registry.settings
+    statement = """
+                SELECT name, revised, authors, abstract,
+                'http://cnx.org/contents/'||
+                ident_hash( uuid, major_version, minor_version) AS link
```
pumazi: It'd be better to use do this url building outside of the query statement. For instance, build the ident_hash in the query (i.e. what you've got without the url prefix concatenation), then on row retrieval use the framework to build the correct url (e.g. `rows = cursor.fetchall(); request.route_url(rows[0][4])`). This will make the url dynamic to whatever environment we put the code in.

pumazi: (updated to use the correct parameters for `request.route_url`)

sisiz: changed to ignore major and minor version because just uuid gives the most recent version anyway. So outside of the sql it does http://cnx.org/contents/ + uuid

cnxarchive/views.py
```diff
@@ -368,6 +368,16 @@ def html_date(datetime):
...
 
 
+def html_rss_date(datetime):
+    """
+    Return the HTTP-date format of python's datetime time.
+
+    Based on:
+    https://legacy.cnx.org/content/recent.rss
+    """
+    return datetime.strftime("%b %d, %Y at %I:%M %p")
```
reedstrm: There's a spec that says this is supposed to be RFC822 (or perhaps RFC2822) formatted. The correct call here would be `strftime("%a, %d %b %Y %H:%M:%S %z")`  (I note that the legacy feed also gets this wrong, but not the same way this code does :-) )

cnxarchive/views.py
```diff
@@ -981,3 +1008,45 @@ def robots(request):
...
+                """
+    with psycopg2.connect(settings[config.CONNECTION_STRING]) as db_connection:
+            with db_connection.cursor() as cursor:
+                cursor.execute(statement,
+                               vars=(portal_type, num_entries, start_entry))
+                search_results = cursor.fetchall()
+    # alter the results to be in the correct form
+    latest_modules = []
+    titles = ["name", "revised", "authors", "abstract", "uuid"]
+    for i in range(len(search_results)):
```
reedstrm: this whole converting results to a dict can be done by using a DictCursor as so:
 `db_connection .cursor(cursor_factory=psycopg2.extras.DictCursor)` then just use search_results (or name it latest_modules and keep the for loop below the same)

cnxarchive/__init__.py
```diff
@@ -71,13 +71,18 @@ def add_route(name, path, *args, **kwargs):
     add_route('legacy-redirect', '/content/{objid}{ignore:(/)?}')  # noqa cnxarchive.views:redirect_legacy_content
     add_route('legacy-redirect-latest', '/content/{objid}/latest{ignore:(/)?}{filename:(.+)?}')  # noqa cnxarchive.views:redirect_legacy_content
     add_route('legacy-redirect-w-version', '/content/{objid}/{objver}{ignore:(/)?}{filename:(.+)?}')  # noqa cnxarchive.views:redirect_legacy_content
+    add_route('recent', '/recent.rss')  # cnxarchive.views:recent
```
reedstrm: a vote on the slack channel came down in favor of `/feeds`

cnxarchive/views.py
reedstrm: Hmm, I need to learn to make my comments on the _last_ line of relevant code, so that the snippets in the discussion feed are easier to read. :-)

sisiz: @reedstrm Is this change necessary? Because when I change it to db_connection.cursor(cursor_factory=psycopg2.extras.DictCursor) and try to cursor.fetchall() it is giving me a list and only returns the class 'psycopg2.extras.DictRow' type when I use fetchone() which then seems like more work to fetch it one at a time

sisiz: @reedstrm When I try to run this with `pserve` I am getting this error: `AttributeError: 'module' object has no attribute 'extras'` even though psycopg2 is imported at the top of the file

cnxarchive/tests/test_views.py
```diff
@@ -3051,3 +3051,23 @@ def patched_now_side_effect(timezone):
...
+        dates = []
+        for module in recent['latest_modules']:
+            dates.append(module["revised"].split(',')[1])
+            keys = module.keys()
+            keys.sort()
+            self.assertEqual(keys, ["abstract", "authors", "name",
+                                    "revised", "uuid"])
+        dates_sorted = list(dates)
+        dates_sorted.sort(reverse=True)
+        self.assertEqual(dates_sorted, dates)
```
reedstrm: FYI this can be spelled: `sorted(dates, reverse=True)`

PR #489, opened 2017-06-13
Oai feed

sisiz:
oai feed implementing the same features as the legacy oai feed. (sets not supported). Also a change was made in cnx-db on this pull request to update the query used for search to take in extra format elements so that query.sql could be reused for the oai-feed and the search. (https://github.com/Connexions/cnx-db/pull/33)
Syntax changes from legacy OAI feed:
-syntax for query in GetRecord changed to match cnx search api (https://github.com/Connexions/cnx-archive/blob/master/docs/search_api_doc.rst)
-weights are comma separated as one value `?weights=title:5,author:2`
- changed syntax `b_start:int=10` to just be `b_start=10` to match rss and search api
- the sort options are only pubdate, version, and popularity because I reused the interface of search.py and those are the sort options there.

### DIFF
--- a/cnxarchive/oai.py
+++ b/cnxarchive/oai.py
@@ -1,0 +1,267 @@
+# -*- coding: utf-8 -*-
+# ###
+# Copyright (c) 2013, Rice University
+# This software is subject to the provisions of the GNU Affero General
+# Public License version 3 (AGPLv3).
+# See LICENCE.txt for details.
+# ###
+import os
+import json
+import logging
+from datetime import datetime, timedelta
+from re import compile, match
+
+import psycopg2
+import psycopg2.extras
+from pyramid import httpexceptions
+from pyramid.settings import asbool
+from pyramid.threadlocal import get_current_registry, get_current_request
+from pyramid.view import view_config
+
+from . import config
+from . import cache
+from . import database
+from .search import (
+    DEFAULT_PER_PAGE, QUERY_TYPES, DEFAULT_QUERY_TYPE, DEFAULT_SEARCH_WEIGHTS,
+    Query, _build_search, QueryResults
+    )
+from .views import html_rss_date
+
+# ################### #
+#     OAI HELPERS     #
+# ################### #
+
+
+def _setFromUntil(request):
+    arguments = []
+    where = ""
+    if 'from' in request.GET.keys():
+        where = "WHERE revised>=(%s)"
+        arguments.append(request.GET.get('from'))
+    if 'until' in request.GET.keys():
+        arguments.append(request.GET.get('until'))
+        if where == "":
+            where = "WHERE revised<=(%s)"
+        else:
+            where += " AND revised<=(%s)"
+    return where, arguments
+
+
+def _databaseDictResults(statement, arguments, settings):
+    with psycopg2.connect(settings[config.CONNECTION_STRING]) as db_c:
+        cur = db_c.cursor(cursor_factory=psycopg2.extras.DictCursor)
+        cur.execute(statement, vars=arguments)
+        search_results = cur.fetchall()
+    return search_results
+
+
+def _parseOaiArguments(verb, request):
+    VERBS = {'GetRecord': {'required': ['verb', 'identifier',
+                                        'metadataPrefix'],
+                           'optional': []},
+             'Identify': {'required': ['verb'],
+                          'optional': []},
+             'ListIdentifiers': {'required': ['verb', 'metadataPrefix'],
+                                 'optional': ['from', 'until', 'set']},
+             'ListMetadataFormats': {'required': ['verb'],
+                                     'optional': ['identifier']},
+             'ListRecords': {'required': ['verb', 'metadataPrefix'],
+                             'optional': ['from', 'until', 'set']},
+             'ListSets': {'required': ['verb'],
+                          'optional': ['resumptionToken']},
+             }
+    # verify that the arguments given are correct, otherwise set an error
+    if not(verb in VERBS.keys()):
+        return {'code': 'badVerb',
+                'message': 'Illegal verb: Legal verbs are: {}'.
+                           format(VERBS.keys())}
+    # check to make sure only correct arguments have been passed
+    all_query_vars = request.GET.keys()
+    required = VERBS[verb]['required']
+    optional = VERBS[verb]['optional']
+    missing = list(set(required) - set(all_query_vars))
+    if missing:
+        return {'code': 'badArgument',
+                'message': 'Required argument {} missing'.format(missing)}
+    extras = list(set(all_query_vars) - (set(required) | set(optional)))
+    if extras:
+        return {'code': 'badArgument',
+                'message': 'Illegal arguments: {}'. format(extras)}
+    return "No errors"
+
+
+def _verifyMetadataPrefix(request):
+    prefixes = ['oai_dc', 'ims1_2_1', 'cnx_dc']
+    prefix = request.GET.get('metadataPrefix')
+    if prefix not in prefixes:
+        return {'error': {'code': 'cannotDisseminateFormat',
+                          'message': 'metadataPrefix {} not supported'.
+                                     format(prefix)}}
+    return {'metadataPrefix': prefix}
+
+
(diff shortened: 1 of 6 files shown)


pumazi:
@sisiz Can you rebase this?

Any chance you could make another PR that focuses on making just the associated changes for Connexions/cnx-db#33 independent of this OAI feed feature? The reason is so that we have small feature changes rather than two or more features in one pull request. This makes it easier to follow and review the changes in the near future and for archaeological usage later.

sisiz:
The one failing test should pass after the cnx-db change is accepted.
https://github.com/Connexions/cnx-db/pull/33

cnxarchive/search.py
```diff
@@ -631,9 +631,48 @@ def _build_search(structured_query, weights):
...
+       ts_headline(ab.html,plainto_tsquery(%(text_terms)s),
+            'ShortWord=5, MinWords=50, MaxWords=60') as abstract,
+       -- until we actually do something with it
+       -- ts_headline(mfti.fulltext, plainto_tsquery(%(text_terms)s),
+       --            'StartSel=<b>, StopSel=</b>, ShortWord=5, MinWords=50,
+       --                MaxWords=60') as headline
+       NULL as headline
+     -- Only retrieve the most recent published modules.
+                  """
+
```
reedstrm: This has been brought back from cnx-db, basically gutting the search query template. Not sure that's the best idea.

cnxarchive/views.py
```diff
@@ -436,6 +436,256 @@ def format_author(personids, settings):
...
+    settings = request.registry.settings
+    search_results = _databaseDictResults(statement, arguments, settings,
+                                          is_dict=False)
+    results = []
+    for result in search_results:
+        results.append(result[0])
+    _formatOaiResults(results, return_vars, request)
+
+
+# ################### #
```
reedstrm: This is a lot of code for the views file. Perhaps his should live in an oai.py, like search does.

cnxarchive/views.py
```diff
@@ -1024,3 +1274,42 @@ def recent(request):
...
+    if verb == 'ListSets':
+        _oaiListSets(return_vars, request)
+    if verb == 'ListIdentifiers':
+        _oaiListIdentifiers(return_vars, request)
+    if verb == 'SearchRecords':
+        _oaiSearchRecords(return_vars, request)
+    elif verb == 'GetRecord':
+        _oaiGetRecords(return_vars, request)
+    elif verb == 'ListRecords':
+        _oaiListRecords(return_vars, request)
```
reedstrm: This is begging for a Dispatcher pattern, where the oai.py would provide a dictionary of the actual execution functions (which traditionally are named do_<action>), keyed on the verb. Also, returning values in an argument (particularly the first argument) is not very pythonic. I'd rather see the parsing and error catch, then the dispatch (which does the actual queries against the db), returning the result dict, which is used to update the results dict:
something like: `result_vars.update(oaiQueries(verb)(request))`

cnxarchive/search.py
sisiz: yeah. I didn't know the best way to do the search because I was gonna use the search build from cnx-db and just change the columns it was selecting, but selecting the columns was a large part of query.sql in cnx-db. But if the search should change then This can be should back.

sisiz:
@reedstrm by dropping the non-standard extension search do you mean change the interface back to how it was on legacy? Or do you mean getting rid of SearchRecord all together because it is not in the standard protocol (https://www.openarchives.org/OAI/openarchivesprotocol.html)

cnxarchive/oai.py
```diff
@@ -0,0 +1,267 @@
...
+    new_vars.update(_formatOaiResults(results, request))
+    return new_vars
+
+
+# ################## #
+#      OAI VIEW      #
+# ################## #
+
+
+@view_config(route_name='oai', request_method='GET',
```
reedstrm: Hmm, kinda thought this part (the actual view) would stay inside views.py

cnxarchive/oai.py
```diff
@@ -0,0 +1,267 @@
...
+                                             'val': request.GET[var]})
+
+    verb = request.GET.get('verb')
+    parse_results = _parseOaiArguments(verb, request)
+    return_vars['verb'] = verb
+    if parse_results != "No errors":
+        return_vars['error'] = parse_results
+        return return_vars
+
+    oaiQuery = {'Identify': do_Identify(request),
```
reedstrm: This dict would stay in oai.py, and be imported from there by views.py for use here.

pumazi:
I'm ok with the view being outside the `views.py` module as long as we have the intention to make one of our next tasks something like, "break up the views module into a views package with specific modules in it." ( I expounded upon this in chat, so not to wiggle away from the context of this PR. )

cnxarchive/templates/oai_verbs.xml
```diff
@@ -0,0 +1,213 @@
...
+                  </ims1_2_1:centity>
+                {% endfor %}
+                <ims1_2_1:date>
+                  <ims1_2_1:datetime>{{record.revised}}</ims1_2_1:datetime>
+                </ims1_2_1:date>
+              </ims1_2_1:contribute>
+            </ims1_2_1:lifecycle>
+            <ims1_2_1:technical>
+              <ims1_2_1:format>text/html</ims1_2_1:format>
+              <ims1_2_1:location type="URI">http://cnx.org/contents/{{record.uuid}}@</ims1_2_1:location>
```
pumazi: How about we derived the url rather than hardcode it?

reedstrm: Ah, template the hostname? Or what do you mean by derived?

sisiz: Yeah, I get what you mean. Will fix that now

sisiz:
@pumazi making the url derived rather than hardcoded. But I was having issues using `request.route_url('content', ident_hash=result['uuid'])` like for the rss feed. But it was brining up this error `ComponentLookupError: (<InterfaceClass pyramid.interfaces.IRoutesMapper>, u'')` so I derived it from the host_url doing the route part manually.

reedstrm:
I was going to merge this, but will wait for @pumazi to respond. :-)

cnxarchive/templates/oai_verbs.xml
```diff
@@ -0,0 +1,213 @@
...
+{% endblock %}
+
+{% block listIdentifiers %}
+  {% for result in results %}
+    <headers>
+      <header tal:condition="r/harvestable | string:True">
+        <identifier>oai:/{{host}}:{{result.uuid}}</identifier>
+        <datestamp>{{result.revised}}</datestamp>
+      </header>
+    <headers>
```
reedstrm: needs to be a `</header>`

cnxarchive/templates/oai_verbs.xml
```diff
@@ -0,0 +1,213 @@
...
+        {% if metadataPrefix == "oai_dc" %}
+          <oai_dc:dc xmlns:oai_dc="http://www.openarchives.org/OAI/2.0/oai_dc/"
+                           xmlns:dc="http://purl.org/dc/elements/1.1/"
+                           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                           xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/oai_dc/
+                           http://www.openarchives.org/OAI/2.0/oai_dc.xsd">
+            <dc:title>{{record.title}}</dc:title>
+            {% for author in record.authors %} <dc:creator>{{author.fullname}}</dc:creator> {% endfor %}
+            {% for subject in record.subjects %} <dc:subject>{{subject}}</dc:subject> {% endfor %}
+            <dc:description>{{record.abstract}}</dc:description>
```
reedstrm: this doesn't handle utf8 abstracts well. (errors)

pumazi:
I'd really like to see us make the `request.route_url` method work. I quickly tried the method from the `pshell` and it worked. In what contextual environment did you do this work, unit tests or within a running application process?

sisiz:
Fixed it to use `request.route_url ` It was that I had removed something from the testing set up that i thought wasn't required.

PR #491, opened 2017-06-21
Organizing views

sisiz:
splitting up the views file into multiple smaller files


sisiz:
Not really sure why the dates are messed up.
locally it was consistently returning the same date(2017-05-16T11:07:37-05:00), just not the date in the assertions originally(2015-03-04T10:03:29-08:00), but on travis it seems to be returning the current date and time
but I think it might have to do with the code around here: https://github.com/Connexions/cnx-archive/blob/organizing_views/cnxarchive/views/exports.py#L114

cnxarchive/views/views_helpers.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
```
pumazi: Any chance we can just call this module `helpers`? Because it's already in the `views` package.

cnxarchive/views/search_view.py
```diff
@@ -0,0 +1,174 @@
+# -*- coding: utf-8 -*-
```
pumazi: This could just be `search`, because it's in the `views` package.

cnxarchive/views/content.py
```diff
@@ -0,0 +1,221 @@
...
+def get_content(request):
+    """Retrieve content using the ident-hash (uuid@version).
+
+    Depending on the HTTP_ACCEPT header return HTML or JSON.
+    """
+    if 'application/xhtml+xml' in request.headers.get('ACCEPT', ''):
+        return get_content_html(request)
+
+    else:
+        return get_content_json(request)
```
reedstrm: [moved my comments about the actual design of the API to @philschatz napkin-notes about  our APIs :-) ]
Ah, you mean content-extras view def should be in contents.py, not extras.py. TBH, it shares code with neither contents nor extras. I'll bow to what you think.

cnxarchive/views/views_helpers.py
reedstrm: +1

cnxarchive/views/content.py
reedstrm: Ah, but its easier to find since the python filename matches the route. Which is does, now.

reedstrm:
I think the last set of changes addressed the most recent request from @pumazi, so merging now,

PR #493, opened 2017-06-27
Extras states

sisiz:
Adding the state to the extras view. Getting the state from the stateid in the modules table for the module with matching uuid and version.

### DIFF
--- a/cnxarchive/views/content.py
+++ b/cnxarchive/views/content.py
@@ -157,9 +157,7 @@ def get_state(cursor, id, version):
     """Determine the state of the content."""
     args = join_ident_hash(id, version)
     sql_statement = """
-    SELECT row_to_json(combined_rows) AS licenses
-    FROM (
-    SELECT m.stateid, ms.statename
+    SELECT ms.statename
     FROM modules as m
     JOIN modulestates as ms
     ON m.stateid=ms.stateid
@@ -167,7 +165,11 @@ def get_state(cursor, id, version):
     """
 
     cursor.execute(sql_statement, vars=(args,))
-    return cursor.fetchall()
+    res = cursor.fetchone()
+    if res is None:
+        return None
+    else:
+        return res[0]
 
 
 def get_export_allowable_types(cursor, exports_dirs, id, version):
--- a/cnxarchive/views/content.py
+++ b/cnxarchive/views/content.py
@@ -153,6 +153,23 @@ def is_latest(id, version):
     return get_latest_version(id) == version
 
 
+def get_state(cursor, id, version):
+    """Determine the state of the content."""
+    args = join_ident_hash(id, version)
+    sql_statement = """
+    SELECT row_to_json(combined_rows) AS licenses
+    FROM (
+    SELECT m.stateid, ms.statename
+    FROM modules as m
+    JOIN modulestates as ms
+    ON m.stateid=ms.stateid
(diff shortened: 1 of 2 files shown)


cnxarchive/views.py
```diff
@@ -115,6 +115,32 @@ def is_latest(id, version):
...
+    JOIN modulestates as ms
+    ON m.stateid=ms.stateid
+    WHERE uuid=%s
+    AND major_version=%s
+    """
+    if len(version_split) == 2:
+        sql_statement += "AND minor_version=%s"
+    sql_statement += ";"
+
+    cursor.execute(sql_statement, args)
```
reedstrm: This can be written "WHERE ident_hash(uuid, major_version, minor_version) = %s" and "args='@'.join((id,version))", and skip the splitting of the version.

cnxarchive/views.py
```diff
@@ -115,6 +115,32 @@ def is_latest(id, version):
...
+    temp = cursor.fetchone()
+    if temp is None:
+        return u'No Stateid in Modules table'
+    else:
+        return temp[0]
```
reedstrm: While commendable to consider error cases, the extra guard code for no return here is not needed - `modules.stateid` has a foreign key relationship with the `modulestates` table, so it's not possible to put a value in that field that does not have a corresponding row in the other table.

reedstrm:
Next step will be an additional `extras` field for the top couple baking results, from the `document_baking_result_associations` table. Not sure if that  should be replacing `state` w/ a structure, or along side it as a separate field `baking_state` perhaps. General state includes a couple other possibilities that aren't directly tied to baking.

cnxarchive/views.py
sisiz: All the stateids in modules match a stateid in modulestates, but some modules don't have a stateid (ie. the stateid field is empty) and that was causing errors

sisiz:
@reedstrm So this PR only adds the state (based on stateid, bot the baking state) to extras. Originally you had wanted to add the baking state as another field which brought up issues of if this belonged in archives or publishing. But has adding the baking status to this been scratched because of the content status view that has the table of the baking status of different books?

cnxarchive/views/content.py
```diff
@@ -153,6 +153,26 @@ def is_latest(id, version):
...
+    FROM modules as m
+    JOIN modulestates as ms
+    ON m.stateid=ms.stateid
+    WHERE ident_hash(uuid, major_version, minor_version) = %s
+    """
+
+    cursor.execute(sql_statement, vars=(args,))
+    res = cursor.fetchone()
+    if res is None:
+        return u'No Stateid in Modules table'
```
pumazi: Let's expose this as `None`. For the most part nothing on the front-end is going to be able to decipher the meaning of this since the terminology is very different on the front-end. Also, `None` or `null` will be more appropriate for javascript logic switching since the other 'state names` will be constant.

kerwinso:
Just tested this on https://archive-qa.cnx.org/extras/185cbf87-c72e-48f5-b51e-f14f21b5eabd@10.115 (Biology). The new `state` field appears. When baking it appears as `processing`, then changes back to `current` when baking succeeds.

image

image

A couple of observations:
1) Clearing varnish cache is required to see updates to the new `state` field.
2) If you rebake a collection, the new `state` does not appear to change at the module level.

Hopefully this is the expected behavior.

PR #496, opened 2017-07-14
Improve mock plpy api compatability

reedstrm:
need this to debug plpy requiring trigger methods


reedstrm:
These test pass locally  - travis breaks inside a python system library, in the generic setup, when deep copying the settings collection. @pumazi any ideas?

PR #497, opened 2017-07-17
Ensure json

reedstrm:
Seem plpy doesn't do as much auto-type conversion as psycopg2, so the simulated plpy in the tests didn't catch this case.

### DIFF
--- a/cnxarchive/tests/test_database.py
+++ b/cnxarchive/tests/test_database.py
@@ -984,6 +984,12 @@ ALTER TABLE modules DISABLE TRIGGER module_published""")
             })
 
     @testing.plpy_connect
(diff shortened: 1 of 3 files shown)


pumazi:
Everything looks good. It's just that the tests aren't passing.

reedstrm:
right - @pumazi any clue what might be causing that? I can't replicate it. The error happens when copying the settings collection during test setup. I'm at a lost to see how this is getting tripped.

pumazi:
This is the same issue that is plaguing the other PRs. I'm looking into it, but having found anything yet.

reedstrm:
I'd like this merge.d The code coverage metric is bogus for the size of the patch, and that it is mock code that's not covered, anyway.

PR #498, opened 2017-07-18
Travis test

reedstrm:
This wrapping as a dict before copy at least passes the test that was previously failing.


pumazi:
I believe this is due to a change between pyramid 1.8 and 1.9. The `pyramid.paster:get_appsettings` function returns a `ConfigDict` object rather than a `dict`.

reedstrm:
Ah, glad you found the root cause. Looks like they made the entire config system/syntax pluggable, and this is fallout from that. So, wrapping it a s a plain dict seems like a reasonable fix.

PR #500, opened 2017-07-20
Oai feed

sisiz:
Alan reviewed the code for QA test and it passed. But Ross had two change requests with the template that I fixed in this PR:

change requests from Ross: Additional acceptance criteria from Ross: - Extra slash on start of Identifier needs to be removed.
- tal & metal attributes should go away.


sisiz:
@Stackblocks Not sure if I was meant to add you as a reviewer because you did the QA test. The issues were just in the templates, not in the real code. But I'm a bit fuzzy on the code review work flow process, so feel free to remove yourself as a reviewer :)

Issue #499, opened 2017-07-19
Need a way to retrieve books that contain a given page


caderitter:
This'll do it (tested, but not probably needs more):

```
CREATE OR REPLACE FUNCTION get_books_containing_page(uuid TEXT, version TEXT) RETURNS SETOF int as $$
  WITH RECURSIVE t(node, title, parent, path, value) AS (
    SELECT nodeid, title, parent_id, ARRAY[nodeid], documentid
    FROM trees tr, modules m
    WHERE m.uuid = $1::uuid
    AND module_version(m.major_version, m.minor_version) = $2
    AND tr.documentid = m.module_ident
    AND tr.parent_id IS NOT NULL
  UNION ALL
    SELECT c1.nodeid, c1.title, c1.parent_id, t.path || ARRAY[c1.nodeid], c1.documentid /* Recursion */
    FROM trees c1
    JOIN t ON (c1.nodeid = t.parent)
    WHERE not nodeid = any (t.path)
  )
  SELECT m.module_ident
  FROM t
  JOIN modules m on t.value = m.module_ident
  WHERE t.parent IS NULL;  	
$$ LANGUAGE SQL;
```

brittweinstein:
CLOSING. If this presents itself as an issue again, we will reopen.

PR #502, opened 2017-07-25
Implement get books containing a given page

caderitter:
Addresses #499. ~~@philschatz mentioned there may be issues with the route `/contents/{uuid}/included-in`, in the case of a page having the title "included-in." Looking for suggestions for a different route name. `/utils/{uuid}/included-in`?~~ Resolved.
Sibling PR [here in cnx-db](https://github.com/Connexions/cnx-db/pull/37), which is needed for this to work.

Edit: touched up in Connexions/cnx-db#79

### DIFF
--- a/cnxarchive/tests/views/test_content.py
+++ b/cnxarchive/tests/views/test_content.py
@@ -855,6 +855,7 @@ INSERT INTO trees (nodeid, parent_id, title, childorder, is_collated)
                          'application/json')
         output['canPublish'].sort()
         self.assertEqual(output, {
+            u'books': [],
             u'downloads': [{
                 u'created': None,
                 u'details': u'PDF file, for viewing content offline and printing.',
@@ -1103,6 +1104,14 @@ INSERT INTO trees (nodeid, parent_id, title, childorder, is_collated)
                          'application/json')
         output['canPublish'].sort()
         self.assertEqual(output, {
+            u'books': [{
+                u'authors': [u'OpenStaxCollege'],
+                u'title': u'College Physics',
+                u'ident_hash': u'e79ffde3-7fb4-4af3-9ec8-df648b391597@7.1'},
+                {
+                u'authors': [u'OpenStaxCollege'],
+                u'title': u'<span style="color:red;">Derived</span> Copy of College <i>Physics</i>',
+                u'ident_hash': u'a733d0d2-de9b-43f9-8aa9-f0895036899e@1.1'}],
             u'state': u'current',
             u'canPublish': [
                 u'OpenStaxCollege',
--- a/cnxarchive/views/content.py
+++ b/cnxarchive/views/content.py
@@ -16,7 +16,6 @@ from pyramid.settings import asbool
 from pyramid.threadlocal import get_current_registry, get_current_request
 from pyramid.view import view_config
 
-from .. import config
 from ..database import (
     SQL, get_tree, get_collated_content, get_module_can_publish, db_connect)
 from ..utils import (
@@ -193,6 +192,22 @@ def get_export_allowable_types(cursor, exports_dirs, id, version):
             pass
 
 
+def get_books_containing_page(uuid, version):
+    """Return a list of book names and UUIDs
+    that contain a given module UUID."""
+
+    with db_connect() as db_connection:
+        with db_connection.cursor() as cursor:
+            cursor.execute(SQL['get-books-containing-page'],
+                           {'document_uuid': uuid,
+                            'document_version': version})
+            results = [{'title': res[0],
+                        'ident_hash': res[1],
+                        'authors': res[2]}
+                       for res in cursor.fetchall()]
+
+    return results
+
 # ######### #
 #   Views   #
 # ######### #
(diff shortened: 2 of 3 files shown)


cnxarchive/views/content.py
```diff
@@ -266,3 +266,41 @@ def get_extra(request):
...
+
+    results = {'books': []}
+    cur_res = []
+
+    with psycopg2.connect(settings[config.CONNECTION_STRING]) as db_connection:
+        with db_connection.cursor() as cursor:
+            cursor.execute(SQL['get-books-containing-page'], {'page_uuid': uuid, 'page_version': version})
+
+            try:
+                cur_res = cursor.fetchall()
```
DarkPrinceFrost: Could save yourself some trouble and loop through the results instead of saving them via a `fetchall`, then iterating afterwards.

cnxarchive/views/content.py
```diff
@@ -266,3 +266,41 @@ def get_extra(request):
     resp.content_type = 'application/json'
     resp.body = json.dumps(results)
     return resp
+
+
+@view_config(route_name='content-included-in', request_method='GET')
+def get_books_containing_page(request):
+    """Return a list of book names and UUIDs that contain a given module UUID."""
```
pumazi: 💯  for including a docstring!

cnxarchive/views/content.py
```diff
@@ -266,3 +266,41 @@ def get_extra(request):
...
+    results = {'books': []}
+    cur_res = []
+
+    with psycopg2.connect(settings[config.CONNECTION_STRING]) as db_connection:
+        with db_connection.cursor() as cursor:
+            cursor.execute(SQL['get-books-containing-page'], {'page_uuid': uuid, 'page_version': version})
+
+            try:
+                cur_res = cursor.fetchall()
+            except (TypeError, IndexError,):  # None returned
```
pumazi: This exception will never be hit. `cursor.fetchall()` will always return something, even an empty list if no rows are found. You'd be better suited to checking with an `if not cur_res: raise ...`.

cnxarchive/views/content.py
```diff
@@ -266,3 +266,41 @@ def get_extra(request):
...
+                raise httpexceptions.HTTPNotFound()
+
+    for res in cur_res:
+        results['books'].append({'name': res[0], 'uuid': res[1]})
+
+    resp = request.response
+    resp.status = "200 OK"
+    resp.content_type = 'application/json'
+    resp.body = json.dumps(results)
+    return resp
```
pumazi: `return results` and add `renderer` in the `@view_config(..., renderer='json')`

caderitter:
@philschatz @pumazi made the requested changes!

pumazi:
@caderitter Looks like you've got some long lines that are not [pep8](https://www.python.org/dev/peps/pep-0008/) compliant. This is preventing travis from reporting an all clear.

caderitter:
@philschatz @reedstrm think i've got everything done!

karenc:
Ready for review again.

philschatz:
Are these related books sorted by when this page was added (that way we prefer the original book over derived copies), or when a book was last updated?

karenc:
Add list of books containing page to page /extras/
There is currently no way to find out which books contain a particular
page other than querying the database directly.  This change adds a list
of books with title, authors and ident hash to the page extras json API.

When user goes to `/extras/{page_ident_hash}`, the json will now include
a list of books, for example:

```
{"books": [
    {"authors": ["OpenStaxCollege"],
     "title": "College Physics",
     "ident_hash": "e79ffde3-7fb4-4af3-9ec8-df648b391597@7.1"},
    {"authors": ["OpenStaxCollege"],
     "title": "<span style="color:red;">Derived</span> Copy of College <i>Physics</i>",
     "ident_hash": "a733d0d2-de9b-43f9-8aa9-f0895036899e@1.1"}],
 "state": "current",
 "canPublish": [
    "OpenStaxCollege",
    "cnxcap"],
 "isLatest": True,
 "downloads": [
    {"created": "2015-03-04T10:03:29-08:00",
     "format": "PDF",
     "details": "PDF file, for viewing content offline and printing.",
     "filename": "useful-inførmation-5.pdf",
     ...
```

RoyEJohnson:
This could be a vital tool for the Highlighting project in Tutor. As it stands, when on a page, Tutor does not have information about what book you're in or what chapters are in that book. That information is required for the Highlighting project.

pumazi:
@RoyEJohnson, it seems like you'd definitively need to know the book you are working on. Discovering that via this API point is not going to give you a definitive and unique result.

reedstrm:
Yes, that's a deficiency on the tutor side that they're looking to fix. (discussion over in tutor-dev)

kerwinso:
Roy said "the Highlighting project should be ok without this PR, so I don't need it. Someone else should review this PR."

cnxarchive/tests/views/test_content.py
```diff
@@ -1103,6 +1104,14 @@ def test_extra_w_utf8_characters(self):
                          'application/json')
         output['canPublish'].sort()
         self.assertEqual(output, {
+            u'books': [{
+                u'authors': [u'OpenStaxCollege'],
```
pumazi: Depending on where this is to be used, the author id is only minimally useful without the human readable information with it (e.g. fullname, title etc.).

As coded it's fine. We'll want to correct this in a followup changeset.

PR #503, opened 2017-07-27
Fix tree_to_json arg type used in transforms

karenc:
In CnxmlToHtmlReferenceResolver, when the SQL function tree_to_json is
used, the argument types are set to (uuid, text), but the type
definition of tree_to_json is actually (text, text).


pumazi:
This will likely affect #504 & #505.

PR #505, opened 2017-07-27
move trigger code to cnx-db

sisiz:
moved trigger code from cnx-archive but still have to move tests which will be done in a separate PR  for clarity.

These tests are not passing because the code has been moved and the imports have been changed to import from cnx-db.triggers instead of the realtive imports from database in this repo. Should pass with the changes from this PR: https://github.com/Connexions/cnx-db/pull/39


sisiz:
The most recent commit (276d9a00931ee444636fbe36baad8178c1544bf5) just changes the .travis file so that it runs on the corresponding branch in cnx-db that are related to these changes. The commit before (0dd882f4559da3a389830045c853a6b783645e24) is the actual change that removes the trigger code.

PR #506, opened 2017-07-28
Xpath extension

caderitter:
[Required sibling PR in cnx-db.](https://github.com/Connexions/cnx-db/pull/40)
Some notes: 

* there is one endpoint - `/xpath`
* specify a book or page uuid and xpath as follows:

    `/xpath?id={uuid}&q={xpath}`

* xpath must only have single quotes, like so:

    `&q=//*[local-name()='definition']`


cnxarchive/views/xpath.py
```diff
@@ -0,0 +1,147 @@
...
+    resp = request.response
+    resp.status = "200 OK"
+    resp.content_type = 'application/xhtml+xml'
+    resp.body = content
+    return resp
+
+
+def tree_to_html_xpath(tree):
+    """Return html list version of book tree."""
+    ul = etree.Element('ul')
```
philschatz: Just wondering if jinja templates should be used. See [views/templates/](https://github.com/Connexions/cnx-archive/tree/master/cnxarchive/views/templates)

cnxarchive/views/xpath.py
```diff
@@ -0,0 +1,147 @@
...
+        a_elm.text = node['title']
+        if node['id'] != 'subcol':
+            if is_first_node:
+                a_elm.set('href', request.route_path(
+                    # FIXME is there a prettier way to do this?
+                    'xpath-book', ident_hash=node['id']) + "?q=" + request.params.get('q', ''))
+            else:
+                a_elm.set('href', request.route_path(
+                    'xpath-page',
+                    ident_hash=node['id']) + "?q=" + request.params.get('q', ''))
```
philschatz: It seems like the generated link contains an additional `?q=${xpath}` querystring field. Could it include a `#${nearest_element_id}` instead?

That way, when someone clicks the link they will be scrolled to the correct part of the Page.

cnxarchive/views/xpath.py
```diff
@@ -0,0 +1,147 @@
...
+                               {'document_uuid': uuid, 'document_version': version, 'xpath_string': xpath_string})
+            except psycopg2.Error as e:
+                exc = httpexceptions.HTTPBadRequest()
+                exc.explanation = e.pgerror
+                raise exc
+
+            for res in cursor.fetchall():
+                results['results'].append(
+                    {
+                        'module_name': res[0],
```
philschatz: I believe `cnx-archive` uses `module` and `collection` internally but I don't think the API exposes those terms. Could this just be `title:` and `uuid:`?

cnxarchive/views/xpath.py
```diff
@@ -0,0 +1,147 @@
...
+    if 'application/xhtml+xml' in request.headers.get('ACCEPT', ''):
+        return xpath_book_html(request)
+    else:
+        return execute_xpath(request, 'xpath')
+
+
+@view_config(route_name='xpath-page', request_method='GET', renderer='json')
+def xpath_page(request):
+    """Given a page UUID (and optional version), returns a JSON object of results, as in xpath_book()"""
+    return execute_xpath(request, 'xpath-module')
```
philschatz: Can this also generate an HTML list of links (like the book results)? (see `#${nearest_element_id}` comment below)

cnxarchive/__init__.py
```diff
@@ -77,6 +77,8 @@ def add_route(name, path, *args, **kwargs):
     add_route('legacy-redirect-w-version', '/content/{objid}/{objver}{ignore:(/)?}{filename:(.+)?}')  # noqa cnxarchive.views:redirect_legacy_content
     add_route('recent', '/feeds/recent.rss')  # cnxarchive.views:recent
     add_route('oai', '/feeds/oai')  # cnxarchive.views:oai
+    add_route('xpath-book', '/xpath/book/{ident_hash}')
+    add_route('xpath-page', '/xpath/page/{ident_hash}')
```
philschatz: Just wondering if these could be combined into 1 route: `/xpath?id=${book_or_page_uuid}`. This might require looking up the `uuid` type (Book/Page) before executing the proper SQL query

cnxarchive/views/xpath.py
caderitter: Certainly could! This is just a modification of the existing tree_to_html code in `/views/content.py`

philschatz:
It seems Travis is complaining about Pep-8 formatting (see Travis log for details)

cnxarchive/views/xpath.py
caderitter: Hm, well, if it doesn't contain the `?q={xpath}`, then I can't pass anything to the route to generate the xpath results from the page. I may not be understanding what you're asking.

cnxarchive/__init__.py
caderitter: Yeah, I think I'll consolidate them. Shouldn't be hard.

cnxarchive/views/xpath.py
philschatz: Oh then maybe not? I'll defer to @pumazi then 😄

pumazi:
I'm working on resolving the test issue.

cnxarchive/views/xpath.py
philschatz: Oh, can you include example HTML in the PR body?

I was expecting something like this:

```html
<html>
  <body>
    <ul>
      <li><a href="/contents/${page_uuid}#${nearest_html_id}">${page_title}</a></li>
      ...
```

That way folks can just click the link and will be scrolled to the element.

For example: https://archive.cnx.org/contents/f1a25413-3af2-4f97-8aac-dd75a16c8aea#fs-id1169738999189

caderitter: Do you mean like [this?](https://archive.cnx.org/contents/185cbf87-c72e-48f5-b51e-f14f21b5eabd%4010.117.html) Perhaps I misunderstood what you were asking for at the beginning. I was aiming at creating an HTML tree that lists links to getting the xpath results from a page. One use case is our intern project - from the frontend, I will retrieve this tree with `xpath=//*[local-name()=\'definition\']`, then allow the users to select links which will return definitions (skipping the backend step, which will be a proxy, passing the request from the frontend to cnx-archive, then parsing the result into Card objects).

caderitter:
@philschatz @reedstrm tests added! everything seems to be working. last problem is the namespacing issue -  `ARRAY[ARRAY['cnx', 'http://cnx.rice.edu/cnxml']]` does not seem to work as a namespace in the function `xpath()`, as it is specified [here](https://github.com/Connexions/devops/wiki/Database-XPath-querying). Seems like none of those URLs resolve.

caderitter:
Travis seems to be failing because it's not using my branch of cnx-db - how should I fix that?

caderitter:
@reedstrm pushed a "skull and crossbones" commit - build succeeds!

philschatz:
hehe, I think Ross was suggesting that you begin the commit message with `:skull_and_crossbones:` 😄

philschatz:
I'm curious, what does the HTML response look like? Can you post some examples, because I am having trouble starting cnx-archive locally.

I'm hoping that it would look something like this:

```
GET /xpath?id={uuid}&q={xpath}

<html>
  <body>
    <ul>
      <li><a href="/contents/{uuid}#id123">{some text, not sure from where, maybe just the Page Title?}</a></li>
      <li><a href="/contents/{uuid}#id234">...</a></li>
    </ul>
  </body>
</html>
```

That way, people can click the link and go directly to the element that matched in the XPath

cnxarchive/views/xpath.py
```diff
@@ -0,0 +1,197 @@
...
+
+def html_listify_xpath(tree, root_ul_element, parent_id=None):
+    """Recursively construct HTML nested list version of book tree, adding
+    hrefs to the page titles which are links to the the API endpoint which runs
+    the xpath on that page CNXML. The original caller should not call this
+    function with the `parent_id` defined.
+
+    """
+    request = get_current_request()
+    querystring = "&q=" + request.params.get('q', '')
```
karenc: Is it possible to use `urllib.urlencode` instead of hand building the query string?  See https://docs.python.org/2.7/library/urllib.html#examples

karenc:
@philschatz It looks like all the pages in the book are returned.

xpath

I think we should only return a page if xpath returns something.

karenc:
@philschatz would this be ok?

xpath

cnxarchive/tests/data/xpath.html
```diff
@@ -0,0 +1,146 @@
+<html xmlns="http://www.w3.org/1999/xhtml">
```
karenc: @philschatz This is how the html looks like.

cnxarchive/views/xpath.py
```diff
@@ -0,0 +1,148 @@
...
+        return xpath_book_html(request, results)
+
+
+def xpath_book_html(request, results):
+    ul = etree.Element('ul')
+    for item in results:
+        li = etree.SubElement(ul, 'li')
+        a = etree.SubElement(li, 'a')
+        a.set('href', '{}?{}'.format(
+            request.route_path('xpath'),
```
philschatz: Can the `<a href="` links point to the destination content rather than an additional query?

From looking at the code it seems that an XPath query on a book will return a list of links to additional XPath queries (`<a href="/xpath?q=...">`) rather than links to the content in the book (`<a href="/contents/{book_uuid}:{page_uuid}#{nearest_element_id}">`)

karenc: I don't think we have that information.  I guess it would have to be separate SQL or xpath queries?

karenc: We're doing the xpath in the database query, this would require getting all the content and parsing it again in python.  It might be ok, I'll test.

karenc: Is this correct? `%(xpath_string)s/ancestor-or-self::*[@id][1]`

philschatz: yes, that should find the nearest id 😄

karenc:
Updated output html, please have a look again :D @philschatz

karenc:
@philschatz from the latest commit:

Fetch individual index.cnxml for book xpath queries
    
Instead of just doing xpath in the sql query, fetch the index.cnxml for
each page that the sql query returns, and do the xpath search in python
so that we can get the closest ancestor id.
    
 I used this for benchmarking:
    
 ```
time wget -O /dev/null 'http://localhost:6500/xpath.html?q=//*[name()=%27link%27]&id=185cbf87-c72e-48f5-b51e-f14f21b5eabd@10.115'
 ```
    
The time it took in the previous version was 0.891 seconds, and this
version took 2.961 seconds.  It is about 3.3 times slower.

philschatz:
While looking at https://github.com/Connexions/oer.exports/pull/2824 I just remembered, is there a way to define several namespace prefixes when making the XPath query?

For example, change https://github.com/Connexions/cnx-archive/pull/506/files#diff-06d46a59f5f508dd27297c1d5d94ea83R78 to be:

```python
NAMESPACES = {
  'cnx':    'http://cnx.rice.edu/cnxml',
  'c':      'http://cnx.rice.edu/cnxml',
  'system': 'http://cnx.rice.edu/system-info',
  'math':   'http://www.w3.org/1998/Math/MathML', 
  'mml':    'http://www.w3.org/1998/Math/MathML', 
  'm':      'http://www.w3.org/1998/Math/MathML',
  'md':     'http://cnx.rice.edu/mdml',
  'qml':    'http://cnx.rice.edu/qml/1.0',
  'bib':    'http://bibtexml.sf.net/',
  'xhtml':  'http://www.w3.org/1999/xhtml',
  'h':      'http://www.w3.org/1999/xhtml',
  'data':   'http://www.w3.org/TR/html5/dom.html#custom-data-attribute',
  'cmlnle': 'http://katalysteducation.org/cmlnle/1.0',
}

for xpath_result in root.xpath(q, namespaces=NAMESPACES):
```


Would that allow people to write `c:list/c:item` instead of the `*[local-name() = 'list']/*[local-name() = 'list']`?

philschatz:
@pumazi could this be unshelved on Monday before the code freeze if time is permitting? 

It has been useful for QA to find out where they should test. For example: https://github.com/Connexions/webview/pull/1584#issuecomment-328195597

karenc:
Ok let me update it.

karenc:
Also added namespaces to https://github.com/Connexions/cnx-db/pull/40

pumazi:
Just need to merge the dependent PR and remove the ☠️  change that uses that dependent branch.

karenc:
Removed :skull_and_crossbones: commit

PR #509, opened 2017-08-03
Fix multiple copies of new version after republish

### LINKED ISSUE #508: Multiple copies of the same minor version of a book in the modules table
This is probably a problem with republishing a book with a minor version update trigger.  When a new version of a page appears, the books that contain that page gets a minor version update.  It appears that we might be inserting a row in the modules table for each page update without changing the minor version.  You can see in the logs below that col12031 was inserted multiple times as version "5.24":

```
karen@qa00:/var/cnx/venvs/publishing/src/cnx-archive$ sudo grep -E '(Trigger fired|col12031|_version:|moduleid:)' /var/log/postgresql/postgresql-9.4-main.log.1
2017-07-28 15:27:48 CDT LOG:  Trigger fired on m58777 
2017-07-28 15:27:49 CDT LOG:  Trigger fired on col12067
        minor_version: 9                              
        major_version: 4                              
        moduleid: col12067                            
2017-07-28 15:27:49 CDT LOG:  Trigger fired on col12074
        minor_version: 14                             
        major_version: 3                              
        moduleid: col12074                            
2017-07-28 15:27:49 CDT LOG:  Trigger fired on col12105
        minor_version: 28                             
        major_version: 1                              
        moduleid: col12105                            
2017-07-28 15:27:49 CDT LOG:  Trigger fired on col11994
        minor_version: 632                            
        major_version: 1                              
        moduleid: col11994                            
2017-07-28 15:27:50 CDT LOG:  Fixed identifier and version for publication at 'd50f6e32-0fda-46ef-a362-9bd36ca7c97d' with the following values: col12031 and 1.5
2017-07-28 15:27:50 CDT LOG:  Trigger fired on col12031
        minor_version: 24                             
        major_version: 5                              
        moduleid: col12031                            
        minor_version: None                           
        major_version: 6                              
        moduleid: m58777                              
2017-07-28 15:29:28 CDT LOG:  Trigger fired on m58615 
2017-07-28 15:29:28 CDT LOG:  Trigger fired on col12067
        minor_version: 9                              
        major_version: 4                              
        moduleid: col12067                            
2017-07-28 15:29:28 CDT LOG:  Trigger fired on col12074
        minor_version: 14                             
        major_version: 3                              
        moduleid: col12074                            
2017-07-28 15:29:28 CDT LOG:  Trigger fired on col12105
        minor_version: 28                             
        major_version: 1                              
        moduleid: col12105                            
2017-07-28 15:29:29 CDT LOG:  Trigger fired on col11994
        minor_version: 632                            
        major_version: 1                              
        moduleid: col11994                            
2017-07-28 15:29:29 CDT LOG:  Fixed identifier and version for publication at 'd50f6e32-0fda-46ef-a362-9bd36ca7c97d' with the following values: col12031 and 1.5
2017-07-28 15:29:29 CDT LOG:  Trigger fired on col12031
        minor_version: 24                             
        major_version: 5                              
        moduleid: col12031                            
        minor_version: None                           
        major_version: 6                              
        moduleid: m58615                              
2017-07-28 15:30:55 CDT LOG:  Trigger fired on m58616 
2017-07-28 15:30:55 CDT LOG:  Trigger fired on col12067
        minor_version: 9                              
        major_version: 4                              
        moduleid: col12067                            
2017-07-28 15:30:55 CDT LOG:  Trigger fired on col12074
        minor_version: 14                             
        major_version: 3                              
        moduleid: col12074                            
2017-07-28 15:30:55 CDT LOG:  Trigger fired on col12105
        minor_version: 28                             
        major_version: 1                              
        moduleid: col12105                            
2017-07-28 15:30:55 CDT LOG:  Trigger fired on col11994
        minor_version: 632                            
        major_version: 1                              
        moduleid: col11994                            
2017-07-28 15:30:55 CDT LOG:  Fixed identifier and version for publication at 'd50f6e32-0fda-46ef-a362-9bd36ca7c97d' with the following values: col12031 and 1.5
2017-07-28 15:30:55 CDT LOG:  Trigger fired on col12031
        minor_version: 24                             
        major_version: 5                              
        moduleid: col12031                            
        minor_version: None                           
        major_version: 5                              
        moduleid: m58616                              
...
```

karenc:
We are seeing multiple copies of the same version in the modules table
after minor version updates.  Several pages are published one after the
other without waiting for the book to finish baking.  This results in
the same minor version being used for republish.

The reason for this seems to be because we use the latest_modules table
when we decide which books to republish.  The latest_modules table may
not actually have the latest versions of the books, because books only
end up in the latest_modules table once baking is finished.

There are 2 fixes to ensure this does not happen in the future:

1. look through the modules table for the max minor version so we would
   never add to the modules table with the same minor version,
2. use the modules table instead of the latest_modules table for
   deciding which books are latest and should be republished.

Close #508

### DIFF
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -187,7 +187,7 @@ def get_collections(module_ident, plpy):
     # latest collection (which may not be the same as what is in
     # latest_modules)
     plan = plpy.prepare('''
-    WITH RECURSIVE t(node, parent, path, document) AS (
+WITH RECURSIVE t(node, parent, path, document) AS (
         SELECT tr.nodeid, tr.parent_id, ARRAY[tr.nodeid], tr.documentid
         FROM trees tr
         WHERE tr.documentid = $1 and tr.is_collated = 'False'
@@ -195,22 +195,24 @@ def get_collections(module_ident, plpy):
         SELECT c.nodeid, c.parent_id, path || ARRAY[c.nodeid], c.documentid
         FROM trees c JOIN t ON (c.nodeid = t.parent)
         WHERE not c.nodeid = ANY(t.path)
-    )
-    SELECT * FROM (
-        SELECT DISTINCT m.module_ident, m.revised,
+    ),
+    latest(module_ident) AS (
+    SELECT module_ident FROM (
+        SELECT m.module_ident, m.revised,
             MAX(m.revised) OVER (PARTITION BY m.uuid) as latest
-        FROM t, modules m
-        WHERE t.parent IS NULL and t.document = m.module_ident
+        FROM  modules m where m.portal_type = 'Collection'
     ) r
     WHERE r.revised = r.latest
-    ORDER BY r.module_ident
+    )
+    SELECT module_ident FROM t, latest
+        WHERE latest.module_ident = t.document
     ''', ('integer',))
     for i in plpy.execute(plan, (module_ident,)):
         yield i['module_ident']
 
 
 def get_subcols(module_ident, plpy):
-    """Get all the collections that the module is part of."""
+    """Get all the sub-collections that the module is part of."""
     plan = plpy.prepare('''
     WITH RECURSIVE t(node, parent, path, document) AS (
         SELECT tr.nodeid, tr.parent_id, ARRAY[tr.nodeid], tr.documentid
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -156,11 +156,18 @@ def get_current_module_ident(moduleid, plpy):
 
 def get_minor_version(module_ident, plpy):
     """Retrieve minor version only given module_ident."""
+    # Make sure to always return the max minor version that is already in the
+    # database, in case the given module_ident is not the latest version
     plan = plpy.prepare('''\
-        SELECT m.minor_version
-            FROM modules m
-            WHERE m.module_ident = $1
-            ORDER BY m.revised DESC''', ('integer',))
+        WITH t AS (
+            SELECT uuid, major_version
+                FROM modules
+                WHERE module_ident = $1
+        )
+        SELECT MAX(m.minor_version) AS minor_version
+            FROM modules m, t
+            WHERE m.uuid = t.uuid AND m.major_version = t.major_version
+        ''', ('integer',))
     results = plpy.execute(plan, (module_ident,), 1)
     return results[0]['minor_version']
 
@@ -176,6 +183,9 @@ def next_version(module_ident, plpy):
 
 def get_collections(module_ident, plpy):
     """Get all the collections that the module is part of."""
+    # Make sure to only return one match per collection and only if it is the
+    # latest collection (which may not be the same as what is in
+    # latest_modules)
     plan = plpy.prepare('''
     WITH RECURSIVE t(node, parent, path, document) AS (
         SELECT tr.nodeid, tr.parent_id, ARRAY[tr.nodeid], tr.documentid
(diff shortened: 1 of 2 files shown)


cnxarchive/database.py
```diff
@@ -186,9 +196,14 @@ def get_collections(module_ident, plpy):
...
-    FROM t JOIN latest_modules m ON (t.document = m.module_ident)
-    WHERE t.parent IS NULL ORDER BY m.module_ident
+    SELECT * FROM (
+        SELECT DISTINCT m.module_ident, m.revised,
+            MAX(m.revised) OVER (PARTITION BY m.uuid) as latest
+        FROM t, modules m
+        WHERE t.parent IS NULL and t.document = m.module_ident
+    ) r
+    WHERE r.revised = r.latest
+    ORDER BY r.module_ident
```
reedstrm: This is not giving the same results as joining against latest_modules (on a quiescent database - no baking going on), when using a non-latest page module_ident. Not sure if that's really an issue, since this code should only ever be called  when a page has been published, and in that case the existing page must, _ipso facto_, be the current latest for that page.

reedstrm: It seems to give the set of latest collections, even for a non-latest page - i.e. that page does not exist in those collections, it only occurs in older versions of those collections.

karenc: This is kind of complicated... this might be different from what you're saying but...  Is it going to return a book that used to have this page but doesn't anymore?

reedstrm: I take that back - it's giving the set of old collections, that that page is in (my test page is from College Physics - latest version is in 86 books, older version is in 204 (all not latest) books)

reedstrm: no, I think what's happening is that the PARTITION OVER doesn't access the whole table, just the subset that matches the WHERE condition - i.e. the collections that this page is in. The newer ones are not present in the subset - PARITION BY doesn't bring in new rows, it only GROUPs what's there (without collapsing the group to  single row, if you get what I mean)

karenc: Yes I get what you're saying... We need to fix it.  :/

reedstrm: ok I think I have it working, though testing on QA is hard - A bunch of "revised" times got set to the same value, somehow, for particular sets of books. I think it's legit - I think all the minor revs for a given set of derived copies of books are created in the same transaction.

reedstrm: The fix is to add yet another modules alias to the sub-select, and use it to get the uuid of the collection-in-question, then get _all_ the modules rows w/ that uuid, and PARITION OVER and MAX() on that one. I'll push once the test pass locally.

karenc: I tried to do something like:

```
with t as (
    select 188037 as documentid union all select 97821 as documentid
), uuids as (
    select uuid from modules m join t on t.documentid = m.module_ident
), max_versions as (
    select max(module_ident) from modules m join uuids on m.uuid = uuids.uuid group by m.uuid
) select * from max_versions;
```

It is kind of working, but let's see how yours look :D

reedstrm: Looks like this:
```
WITH RECURSIVE t(node, parent, path, document) AS (
        SELECT tr.nodeid, tr.parent_id, ARRAY[tr.nodeid], tr.documentid
        FROM trees tr
        WHERE tr.documentid = $1 and tr.is_collated = 'False'
    UNION ALL
        SELECT c.nodeid, c.parent_id, path || ARRAY[c.nodeid], c.documentid
        FROM trees c JOIN t ON (c.nodeid = t.parent)
        WHERE not c.nodeid = ANY(t.path)
    )
    SELECT module_ident FROM (
        SELECT m.module_ident, m.revised,
            MAX(m.revised) OVER (PARTITION BY m.uuid) as latest
        FROM t, modules m, modules m2
        WHERE t.parent IS NULL
           and t.document = m2.module_ident
           and m2.uuid = m.uuid

    ) r
    WHERE r.revised = r.latest
    ORDER BY r.module_ident
```

karenc: Oooh, I think I see how it's going to work.  Cool!

reedstrm:
@karenc - take a look at my change - I moved your subquery up to make it a second CTE. Seems to do the job.

karenc:
@reedstrm I think it looks good, do you want to `git rebase --autosquash -i` to get rid of that `fixup!` commit?

PR #511, opened 2017-08-04
Set up travis to run on precise

karenc:
Travis is switching from Precise to Trusty as their default Linux environment.
For some reason, we are getting "InternalError: ImportError: No module named
cnxarchive.database" when our tests run on Trusty.

https://travis-ci.org/Connexions/cnx-archive/builds/260765357

This change explicitly tells Travis to use Precise so that our tests will
continue to work while we look into the problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: precise
 language: python
 python:
   - "2.7"


karenc:
Same as https://github.com/Connexions/cnx-publishing/pull/167

PR #520, opened 2017-08-26
Move triggers tests and tests from cnx-archive

sisiz:
Moving the triggers and test cases from cnx-archive to cnx-db

related to https://github.com/Connexions/cnx-db/pull/44


edwoodward:
Opened issue so we can work on this again in the future. Closing PR

https://github.com/Connexions/cnx-archive/issues/607

PR #523, opened 2017-09-12
Add rhaptos.cnxmlutils version to index.cnxml.html

karenc:
When using the rhaptos.cnxmlutils xsl stylesheet on index.cnxml, pass in
rhaptos.cnxmlutils version so it appears in the `<body>` tag in
index.cnxml.html.

Extra code was necessary for rhaptos.cnxmlutils to return its version
when `git` is required.  By default, postgres python triggers have
nothing in $PATH, so versioneer was returning "0+unknown" as the version
number for rhaptos.cnxmlutils.  We need to add `/usr/bin` to $PATH for
versioneer to work correctly.

### DIFF
--- a/cnxarchive/transforms/converters.py
+++ b/cnxarchive/transforms/converters.py
@@ -8,6 +8,7 @@
 """CNXML <-> HTML conversion code."""
 
 import os
+import sys
 from io import BytesIO
 
 import rhaptos.cnxmlutils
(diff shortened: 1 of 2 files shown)


karenc:
The code I added for getting triggers to work is not tested in cnx-archive... :/ causing the coverage to drop significantly.

reedstrm:
Hmm, this only applies for non-egg deployments of `cnx-archive`, so basically development? Won't packaged releases return whatever has been stored as `./_version.py` (hard coded json, looks like)?
So dev environments would get wonky version strings.

cnxarchive/transforms/converters.py
```diff
@@ -29,7 +30,19 @@
...
 def _gen_xsl(f, d=XSL_DIR):
-    return etree.XSLT(etree.parse(os.path.join(d, f)))
+    transform = etree.XSLT(etree.parse(os.path.join(d, f)))
+
+    def transform_w_version(*args, **kwargs):
+        # If git is not in $PATH, rhaptos.cnxmlutils.__version__ returns
+        # "0+unknown"
+        if '/usr/bin' not in os.getenv('PATH', ''):
+            os.environ['PATH'] = '/usr/bin:{}'.format(os.environ['PATH'])
+            reload(rhaptos.cnxmlutils)
```
pumazi: Interesting touch. 😸

karenc: I read that this only works in python2... :p Will need to change it when we move to python3.

karenc:
@reedstrm You're right.  I just checked cnx-db 0.9.0 and _version.py has some json.

karenc:
@reedstrm So what should we do about this?  Not put in a version if it's on QA and TEA? (They use rhaptos.cnxmlutils master and need to use versioneer)  It seems useful to have the version :thinking:

reedstrm:
Yeah, it's worth the pain to have it on the dev instances (esp. the not-really-dev-pretending-to-be-production that is tea) I'm willing to ignore the coverage since it's such a small patch - two lines in one corner case makes the patch coverage percentage ridiculous.

reedstrm:
Yeah, I'm banging up a simple test. Will merge in about 30 min.

cnxarchive/transforms/converters.py
reedstrm: Well, I just re-used it in the test for this feature, so two places to fix it for py3 :-(

karenc:
Thanks!  Looks good

reedstrm:
local run of coverage says my attempt worked - that condition is now covered. We'll see what travis thinks.

reedstrm:
We're all green - merging

PR #528, opened 2017-09-25
Explicitly close all psycopg2 db connections

karenc:
The psycopg2 db connection `with` statement does not actually close the
database connection causing lots of idle postgres connections on the
production servers.  An idle postgres connection can be found by doing
`ps aux` on the database servers, for example you will see something
like this:

```
postgres 25498  0.0  0.5 302652 10996 ?        Ss   09:31   0:00 postgres: cnxarchive cnxarchive 192.168.122.250(42646) idle
```

On http://initd.org/psycopg/docs/usage.html#with-statement

"Note that, unlike file objects or other resources, exiting the
connection’s with block doesn't close the connection but only the
transaction associated with it"

This change creates a `db_connect` function that is like
`psycopg2.connect` but closes the db connection when exiting the `with`
block.

### DIFF
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -6,6 +6,7 @@
 # See LICENCE.txt for details.
 # ###
 """Database models and utilities."""
+import contextlib
 import os
 import json
 import psycopg2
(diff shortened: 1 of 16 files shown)


cnxarchive/database.py
```diff
@@ -73,12 +74,18 @@ def _read_sql_file(name):
...
     if connection_string is None:
         settings = get_current_registry().settings
         connection_string = settings[config.CONNECTION_STRING]
-    return psycopg2.connect(connection_string)
+    db_conn = psycopg2.connect(connection_string)
+    try:
+        with db_conn:
+            yield db_conn
+    finally:
+        db_conn.close()
```
reedstrm: This is probably good for now, but it occurs to me that it provides for the possibility of doing some connection pooling here.

PR #530, opened 2017-10-12
Remove plpy_connect testing implementation in favor of fake_plpy

pumazi:
Remove `plpy_connect` testing implementation in favor or `fake_plpy`. This was only used in two tests. Looks like the later `fake_plpy` implementation one ups it.


pumazi:
Test failures aside... To be completely honest, I like the one I just deleted more than the fake_plpy, because it doesn't assume a global access to the connection factory.

karenc:
Ah, it's also possibly more correct?  `fake_plpy` is creating a new database connection every time `execute` is called.  We can add an issue to follow up?

pumazi:
I'd like to agree that using `PREPARE` and `EXECUTE` is the way to go. But in my work to move the transforms and their tests to cnx-db, I've run into an issue. Apparently `plpy.prepare` does work differently from what happens in a `PREPARE`, because it's able to magically understand that a UUID is not Text, but a UUID type. If we were to switch this around, the existing tests would break.

```
E       ProgrammingError: operator does not exist: text = uuid
E       LINE 2: WHERE uuid::text = $1 and module_version(major_version, mino...
E                                ^
E       HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

reedstrm:
Hmm, can I see the full context for that? The ::text was cast was added explicitly in at least one place like that to avoid erroring when a non-uuid was passed in, so the expectation was that $1 is "unknown" or "text" at this point.

edwoodward:
An issue has been created for this so the PR can be closed and worked on again later.

https://github.com/Connexions/cnx-archive/issues/608

PR #531, opened 2017-10-13
Fix broken tests in cnx-archive master following cnx-db updates

karenc:
- Fix update latest trigger tests to use legacy version in inserts

  We have changed to updating latest_modules by comparing the major and
  minor versions instead of using "revised".  The update latest trigger
  tests insert legacy versions (`1.1`, `1.1`, `1.1`) and major versions
  (`1`, `3`, `2`).  The major versions were actually getting overwritten
  by the legacy compatible triggers, causing all the modules to have the
  major version "1".  Change the test to use correct legacy versions
  `1.1`, `1.3`, `1.2`.

- Install tzdata for cnx-archive docker image

  The ubuntu 16.04 docker image does not have tzdata installed.  So when
  we need to use timezones, it just returns the time in UTC.
  
  ```
  root@e9d0eb00903e:/# date
  Fri Oct 13 01:04:05 UTC 2017
  root@e9d0eb00903e:/# TZ=Europe/London date
  Fri Oct 13 01:04:09 Europe 2017
  root@e9d0eb00903e:/# tzselect
  /usr/bin/tzselect: line 180: /usr/share/zoneinfo/iso3166.tab: No such file or directory
  /usr/bin/tzselect: time zone files are not set up correctly
  ```
  
  Add `tzdata` to the list of packages to install using `apt-get install`.

- Update book search test following changes in book search sql

  The book search sql has been changed in cnx-db to search OR terms,
  rather than AND them.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ RUN apt-get update && apt-get install -y \
     gcc \
     zlib1g-dev \
     wget \
+    tzdata \
  && rm -rf /var/lib/apt/lists/*
 
 COPY . /src/
--- a/cnxarchive/tests/test_database.py
+++ b/cnxarchive/tests/test_database.py
@@ -1722,7 +1722,7 @@ class UpdateLatestTriggerTestCase(unittest.TestCase):
         module_ident, uuid = cursor.fetchone()
 
         cursor.execute('''INSERT INTO modules VALUES (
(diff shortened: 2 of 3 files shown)


karenc:
Since these are unrelated changes, I think it's best if we `Rebase and merge` instead of "Squash and merge".

PR #534, opened 2017-10-25
Change utils and database code to be more py3 compatible

karenc:
- Change dict .iteritems() to .items() in database.py

  Python 3 dicts don't have `.iteritems()` anymore, changing to `.items()`
  means that the code will work in python 2 and 3.  Since cnx-db tests
  are also running in python 3.5, cnxarchive needs to have some of the
  database code updated to be python 3 compatible.

- Change uuid `.get_bytes()` to `.bytes` in utils

  Python 3 uuid objects no longer has `.get_bytes()` but it seems `.bytes`
  does the same thing and exists in python 2 and 3, so changing the code
  to use that instead.

- Define basestring for python 3 in utils ident_hash

  `basestring` is `str` and `unicode` in python 2, the equivalent for
  python 3 would be `bytes` and `str`.

---

Created for Connexions/cnx-db#86

### DIFF
--- a/cnxarchive/database.py
+++ b/cnxarchive/database.py
@@ -468,7 +468,7 @@ def republish_module_trigger(plpy, td):
     plpy.log('modified: {}'.format(modified))
     plpy.log('insert values:\n{}\n'.format('\n'.join([
         '{}: {}'.format(key, value)
-        for key, value in td['new'].iteritems()])))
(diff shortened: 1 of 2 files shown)


karenc:
[Refactor In-Database Logic](https://trello.com/c/OhfwIMfT/231-refactor-in-database-logic)

PR #535, opened 2017-10-30
Add sitemap_index.xml and have separate sitemap files

karenc:
- Rewrite sitemap.py using lxml.etree

  `sitemap.xml` was generated by hand, with an escape function in utils to
  entity escape of urls.  Since `lxml.etree` is already used in other
  parts of cnxarchive, we might as well make use of it and get rid of the
  escape function in utils.  This will also make sure the sitemap.xml is
  valid xml.

- Remove subcollections from sitemap.xml

  Subcollections are currently included in sitemap.xml, but the urls
  return 404, so we probably shouldn't include subcollections in
  sitemap.xml.

- Add from ident to sitemap: `sitemap-{from_id}.xml`

  There is a limit of 50,000 urls in a sitemap xml according to
  sitemaps.org.  To implement split sitemap files, we need to know which
  modules to include, so we are including this in the sitemap url.  These
  sitemap urls are going to be included in the sitemap_index.xml.

- Add sitemap_index.xml and split sitemap files

  There is a limit of 50,000 urls in a sitemap xml according to
  sitemaps.org.  To implement split sitemap files, we need to know which
  modules to include, so we are including this in the sitemap url.  These
  sitemap urls are then included in the sitemap_index.xml.

- Change sitemap url in robots.txt to sitemap_index.xml

  We split sitemap.xml into separate files so it doesn't go over the
  50,000 URLs limit.  The individual files are referenced in
  sitemap_index.xml and it is used as the canonical sitemap url, so we
  need to change the sitemap url in robots.txt.

See also https://github.com/Connexions/cnx-deploy/pull/414


karenc:
[Split sitemap.xml into multiple files](https://trello.com/c/Ry2X5UH7/288-split-sitemapxml-into-multiple-files)

cnxarchive/views/sitemap.py
```diff
@@ -77,3 +81,32 @@ def sitemap(request):
...
+                FROM (
+                    SELECT module_ident,
+                           row_number() OVER (ORDER BY module_ident) AS row
+                    FROM latest_modules
+                    WHERE portal_type NOT IN (
+                        'CompositeModule', 'SubCollection')
+                ) AS t
+                WHERE t.row % {} = 1""".format(SITEMAP_LIMIT))
+
+            for module_ident, _ in cursor.fetchall():
```
reedstrm: If you're just going to ignore the row number, don't return it.

cnxarchive/views/sitemap.py
```diff
@@ -77,3 +81,32 @@ def sitemap(request):
...
+            cursor.execute("""\
+                SELECT t.*
+                FROM (
+                    SELECT module_ident,
+                           row_number() OVER (ORDER BY module_ident) AS row
+                    FROM latest_modules
+                    WHERE portal_type NOT IN (
+                        'CompositeModule', 'SubCollection')
+                ) AS t
+                WHERE t.row % {} = 1""".format(SITEMAP_LIMIT))
```
reedstrm: This could have been written as a CTE (`WITH t(module_ident, row) AS ( …`) which is usually a bit clearer, though in this case, it's pretty much equivalent, in readability and performance. Usually not a fan of `*` in production code, since it can hide bugs when a table changes, but this is a defined subquery, so changes would have to be right here, anyway, so that gets a pass. :-) Ah, but I see we don't use all the values, so just return the module_ident, please.

cnxarchive/tests/test_sitemap.py
```diff
@@ -61,7 +104,9 @@ def test_updatefreq_urlEntry(self):
         """test valid changefreq"""
         ue = sitemap.UrlEntry(loc='http://example.com', changefreq='weekly')
         self.assertEqual(repr(ue), "<UrlEntry 'http://example.com'>")
-        self.assertEqual(str(ue), '<url>\n<loc>http://example.com</loc>\n<changefreq>weekly</changefreq>\n</url>\n')
+        self.assertEqual(
+            str(ue), '<url>\n  <loc>http://example.com</loc>\n'
+                     '  <changefreq>weekly</changefreq>\n</url>\n')
```
reedstrm: Gee, it'd be nice to compute an actual likely value for this. TODO, I suppose.

cnxarchive/views/sitemap.py
```diff
@@ -49,21 +53,21 @@ def notblocked(page):
...
+                SELECT
+                    ident_hash(uuid, major_version, minor_version)
+                        AS idver,
+                    REGEXP_REPLACE(TRIM(REGEXP_REPLACE(LOWER(name),
+                        '[^0-9a-z]+', ' ', 'g')), ' +', '-', 'g'),
+                    revised
+                FROM latest_modules
+                WHERE portal_type NOT IN ('CompositeModule', 'SubCollection')
+                  AND module_ident >= (SELECT COALESCE(%s, 0))
+                ORDER BY module_ident
```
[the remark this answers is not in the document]
reedstrm: I hunted around a bit, and it's not clear that there _is_ utility in this, but I also recall the original intern doing some research and thinking there was.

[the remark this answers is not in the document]
karenc: We can have `sitemap-books-{id}.xml`.

I changed the order because I thought perhaps the old sitemaps can be cached longer, but actually that's wrong... since this is the latest_modules, it's going to change.

cnxarchive/tests/test_sitemap.py
karenc: I actually thought it should be `never`, the urls in our sitemap.xml have versions in them right?

cnxarchive/views/sitemap.py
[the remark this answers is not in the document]
karenc: And the reason I changed `revised` to `module_ident` is, so we can do this in the second sitemap xml easily instead of doing something like `LIMIT 50000 OFFSET 50000` which is apparently very slow compared to something like `module_ident >= 5000`.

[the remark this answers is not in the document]
reedstrm: I'm fine with the numbered sitemaps, and using module_ident. Cacheability is not going to be an issue - these aren't high traffic pages - google/other spyders fetch them maybe once a week.
I think it's not a big deal either way. :)

cnxarchive/views/sitemap.py
reedstrm: That was confusing -  the only call for action in there is to get rid of the ignored `_` return
This is just a general cleanup - avoids an message from flake8 for me, but could go out  as is.

cnxarchive/views/sitemap.py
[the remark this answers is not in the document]
karenc: Ok great!  I'll change the order to use `DESC` again, but keep using `module_ident`

cnxarchive/views/sitemap.py
karenc: That's done.

cnxarchive/tests/test_sitemap.py
karenc: What do you think about changing the `changefreq` to `never`?

cnxarchive/views/sitemap.py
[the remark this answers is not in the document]
karenc: So it sounds like I don't need to change the order?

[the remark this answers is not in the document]
reedstrm: I think the order probably has no affect, so we can leave it.

cnxarchive/tests/test_sitemap.py
karenc: Oh then does it mean we should remove the versions from the urls?

reedstrm: I think we need an actual SEO expert to think about these things. :/

PR #536, opened 2017-10-31
Remove robots.txt (moved to cnx-deploy)

karenc:
DevOps prefer to have robots.txt in cnx-deploy so they can configure it.

See https://github.com/Connexions/cnx-deploy/pull/417


pumazi:
Waiting on the acceptance of connexions/cnx-deploy#417 before merging

kerwinso:
https://github.com/Connexions/cnx-deploy/pull/417 was accepted.

PR #538, opened 2017-11-07
Move transforms to cnx-db

pumazi:
These changes are in response to Connexions/cnx-db#86. This removes the transforms from this package and places them in cnx-db.

PR #539, opened 2017-11-07
Wait for the archive container to be up in .travis.yml

karenc:
There appears to be a race condition on travis:

```
$ docker exec cnxarchive_archive_1 /bin/bash -c "pip install --user coverage codecov"
Error response from daemon: Container 90b9f01437b8f1585fb315d15e0c0979678b943152fd6e1e8edf7b0149303d7c is not running
```

So add a `while sleep` loop to wait for the container to be up before trying to
execute stuff on it.


karenc:
[Move transforms trigger logic to cnx-db](https://trello.com/c/PIxcba94/298-move-transforms-trigger-logic-to-cnx-db)

karenc:
#538 needs this

pumazi:
Seems to work! Good shell skills. 💯

PR #541, opened 2017-11-13
Adjust travis config to give cnxarchive role login perms

pumazi:
I'm not sure why this is necessary now when it wasn't before, but it does make it so the master is passing tests again.


pumazi:
Karen looked at this a little deeper. She'll be putting together a PR to change the `CREATE ROLE` lines to `CREATE USER`.

PR #542, opened 2017-11-24
cache based on baked status for /contents/*

sisiz:
change the `Cache-Control` header based on the baked status. Both for html and json. If it is baked then it is set to no-cache

### DIFF
--- a/cnxarchive/views/content.py
+++ b/cnxarchive/views/content.py
@@ -224,7 +224,7 @@ def get_content_json(request):
     if not(re.compile(r'\w{8}-\w{4}-\w{4}-\w{4}-\w{12}').search(request.url)):
         resp.cache_control = "public"
     else:  # otherwise (there is an ident-hash) then check baking status
-        if (result['baked']) and (result['state'] in [1, 8]):
+        if (result['baked']) and (result['stateid'] in [1, 8]):
             # state 1 = current, state 8 = fallback
             resp.cache_control = "public"
         else:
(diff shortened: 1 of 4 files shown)


cnxarchive/views/helpers.py
```diff
@@ -66,6 +66,14 @@ def get_content_metadata(id, version, cursor):
         #       Until then we will do the replacement here.
         result['mediaType'] = portaltype_to_mimetype(result['mediaType'])
 
+        cursor.execute("""
```
reedstrm: Rather than a second fetch here, let's do a linked change over in cnx-db, to return baked status in `get-module-metadata`

cnxarchive/views/content.py
```diff
@@ -241,6 +244,8 @@ def get_content_html(request):
     resp.status = "200 OK"
     resp.content_type = 'application/xhtml+xml'
     resp.body = content
+    if result['baked'] is not None:
+        resp.cache_control = "no-cache"
```
reedstrm: Ditto. Also, on deeper look at the pyramid `response` object API, let's use `resp.cache_expires(0)` instead of setting the header directly. This will have nice affects for using the http_cache machinery later to set values for caching via config. https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/webob.html?highlight=response.cache_expires

cnxarchive/views/content.py
```diff
@@ -218,8 +218,11 @@ def get_content_json(request):
     """Retrieve content as JSON using the ident-hash (uuid@version)."""
     result = _get_content_json()
 
-    result = json.dumps(result)
     resp = request.response
+    if result['baked'] is not None:
+        resp.cache_control = "no-cache"
```
reedstrm: You're right this logic is backwards. We want to `no-cache` when not baked. Note that `baked` will come across as a `datetime.datetime` object, when set, so the correct test is in fact for `is None`

reedstrm:
@pumazi comments re: pyramid caching?

sisiz:
cnx-db change to remove double fetch: https://github.com/Connexions/cnx-db/pull/114

sisiz:
closed, changes for cache moved to https://github.com/Connexions/cnx-archive/pull/545 which also changes the content routes

PR #544, opened 2018-01-26
Fix travis tests from modification to the cnx-db container

pumazi:
We modified the cnx-db container so that it now requires the `POSTGRES_DB` environment variable be defined when the database isn't `repository`, which it isn't in the case of tests within this repository.

The quotations around the environment variables are because it didn't work otherwise. I didn't look into it to much, but it's likely because the value has a `-` in it that flubs up the yaml parser.

### DIFF
--- a/.travis-compose.yml
+++ b/.travis-compose.yml
@@ -17,10 +17,10 @@ services:
     ports:
       - "5433:5432"
     environment:
-      - DB_USER=cnxarchive
-      - DB_NAME=cnxarchive-testing
-      - DB_URL=postgresql://cnxarchive@localhost/cnxarchive-testing
-      - DB_SUPER_URL=postgresql://postgres@localhost/cnxarchive-testing
+      - "DB_USER=cnxarchive"
+      - "POSTGRES_DB=cnxarchive-testing"
+      - "DB_URL=postgresql://cnxarchive@localhost/cnxarchive-testing"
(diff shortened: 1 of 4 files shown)


pumazi:
The tests are failing on the modified sql statement added in cnx-db==1.5.0. If you need these tests to pass, pin to a cnx-db==1.4.0.

reedstrm:
I'm updating the tests

reedstrm:
There's also this:
$ docker-compose down
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.db.volumes contains an invalid type, it should be an array

karenc:
Oh in .travis.yml, I guess instead of just `docker-compose down`, we need:

```diff
diff --git a/.travis.yml b/.travis.yml
index 9428e6c..1332a68 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
   # This is the same as `python -m unittest discover` with a coverage wrapper.
   #- coverage run --source=cnxarchive setup.py test
   - docker exec cnxarchive_archive_1 /bin/bash -c "cd /src; TESTING_CONFIG=cnxarchive/tests/.travis-testing.ini coverage run --source=cnxarchive setup.py test"
-  - docker-compose down
+  - docker-compose -f .travis-compose.yml down
 after_success:
   # Report test coverage to codecov.io
   - docker exec cnxarchive_archive_1 /bin/bash -c "codecov"
```

PR #545, opened 2018-02-09
One content route and cache control on that route

reedstrm:
This code merges the content* routes, so we can handle cache-control in one place.

### DIFF
--- a/cnxarchive/__init__.py
+++ b/cnxarchive/__init__.py
@@ -60,7 +60,7 @@ def declare_api_routes(config):
         return config.add_route(name, path, *args,
                                 pregenerator=pregenerator(path), **kwargs)
 
-    add_route('content', '/contents/{ident_hash:([^:/@.]+(@[0-9.]*[0-9]+)?)}{separator:(:?)}{page_ident_hash:([^:/@.]+(@[0-9.]*[0-9]+)?)?}{ext:([.](html|json))?}{ignore:(/.*)?}')  # noqa cnxarchive.views:get_content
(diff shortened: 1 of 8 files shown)


pumazi:
I can't say I'm in agreement with this changeset. Instead of letting the framework do the work, we've taken it upon ourselves to do routing logic in the view.

reedstrm:
I'd say rather than hiding complex routing logic in the framework in three different methods across two files, requiring comparing 3 regexes by eye, we're drying out route implementation to allow using the framework for view configuration  - i.e. default cache behavior, and placing the not-obvious routing and cache-override logic on one place with comments.

TBH, I've never like the magic-file-extension-to-change-returned-data routes anyway: HTTP  has a way to specify wanting different types: the accepted-types header. But we're using it, so leaving in place.

reedstrm:
Seems I had the regex slightly wrong. Kudos to everyone who's contributed to the extensive set of route and content tests: caught a couple coding errors here.

reedstrm:
@pumazi Let me clarify the comment re: doing routing in the logic. Currently, there are _two_ ways to select html vs json return type: one is to set an `Accept:` header, that we processed in the `content` route and view: the other is to append ".html" or ".json", which then is matched to either `content-html` or `content-json` views. So there are 3 views associated with 3 routes to serve two different `Content-Type` returns. Which means to do anything with caching consistently across all `/contents` urls requires repeating something three times. This PR collapses the 3 routes to 1, and combines the `Content-Type` selection logic that was otherwise spread between 3 views into one place. It then applies a default caching policy via the framework tools  (view_config parameter) and uses logic in that one view to override for content-specific reasons (baking-in-flight-or-errored). It was while repeating that caching override logic in two places (and about to add the third place) that drove me to do this refactor, instead.

PR #546, opened 2018-02-15
WIP: Changed the short id redirect from HTTPFound to HTTPMovedPermanently

m1yag1:
It was discovered that Google Search may not be indexing the pages on cnx due to the code we are using when redirecting from the legacy uuid to to the short id. This fix makes a change to the code from 302 to 301 in order to increase our googlebot friendliness.


reedstrm:
BTW, @pumazi all the travis builds for cnx-archive have been broken by docker issues, for some time.

reedstrm:
All tests pass locally, so we're going to merge and tag this to get it into the upcoming release, for critsit service.

PR #547, opened 2018-02-19
Restore a /robots.txt route. Returns do not index robots.txt

reedstrm:
As an API server, it isn't needed or proper for search engines to index the archive.cnx.org urls out there. Adding a default "don't index me" route, so that no matter how archive is deployed, we'll fall back to this control file. Note that in production, the  top level human-facing urls robots.txt file is contructed at deployment time, and served as a static file.

### DIFF
--- a/cnxarchive/__init__.py
+++ b/cnxarchive/__init__.py
@@ -72,6 +72,7 @@ def declare_api_routes(config):
     add_route('extras', '/extras')  # cnxarchive.views:extras
     add_route('sitemap-index', '/sitemap_index.xml')  # noqa cnxarchive.views:sitemap
     add_route('sitemap', '/sitemap-{from_id}.xml')  # noqa cnxarchive.views:sitemap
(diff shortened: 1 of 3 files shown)


kerwinso:
Trello card: https://trello.com/c/UbOReNMF

pumazi:
I thought we wanted Ops to manage the robots.txt. For the time being this looks fine though.

Can we do a followup issue to remove this in favor of putting a robots.txt for archive in cnx-deploy? (CC @dtwilliamson @scb6 @DarkPrinceFrost )

reedstrm:
We have extended robots.txt deployment in cnx-deploy. This is a belt-and-suspenders sort of thing to make sure that, as an API server, we never ever want to allow it to be indexed. Alternative implementation would be to extend the routing configs to explicitly recognize archive*/robots.txt and send it to nginx to serve a static file, like we do now for <primary interface>/robots.txt

PR #548, opened 2018-02-20
WIP: Updated db service in docker-compose to use openstax cnx-db image

m1yag1:
* The internal port binding was not working when using the build file from the git repository so using the image may be a better way to do this.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       - db
       - memcached
   db:
-    build: https://github.com/Connexions/cnx-db.git
+    image: openstax/cnx-db:1.5.1
     ports:
       - "5432"
 #    volumes:


pumazi:
This corrects the travis issue from happening again. But it means that we'll need to continuously bump the version when a cnx-db release is made.

The issue was that the base postgres image, which cnx-db is based on, became out of date (this happened once before). The problem is that the base image has an outdated package that cause our Dockerfile instructions to upgrade. This time it wasn't as obvious that an upgrade was the issue, but I'm fairly sure that was the case. I've since added added a new step to [the cnx-db docker image release steps](https://github.com/Connexions/cnx-db/wiki/Release) to always rebuild the base postgres image as well (aka openstax/postgres).

So, it's possible for this issue to happen again if we leave things as they are. Or we can correct them by pinning the container, which adds a period maintenance step. The use of a pinned container does seem to be working well for [Press](https://github.com/Connexions/cnx-press). But Press is different because it is the authority on its dependency pinnings, where cnx-deploy is the authority on the pinnings for cnx-archive, which means we'd be tracking the change in two different places.

reedstrm:
The auto-upgrade of packages in our travis scripts is both a blessing and a curse: it helps push us to upgrade underlying packages, but it causes issues like this. Also, tests take longer and longer to run. I think perhaps the right answer is in fact to pin dependencies in our test infrastructure, like this, and use external tools to track dependencies, like we do for the python libs

PR #549, opened 2018-03-02
Pin the cnx-db docker container version and fix a test

pumazi:
The pinning seen here is similar to #548. See the discussion on that issue for details.

The test fix is because cnx-db has changed since the last successful test run. (We should also think about pinning against a released version of cnx-db, but one step at a time. This is something that is being done in Press and is working well.)

PR #551, opened 2018-03-15
New fields to extras view

sisiz:
Returning new fields in extras view. (authors with formatting, revision date, and short id for each book a page is in). Changes made to be used with webview page to view a page that is navigated to outside of a book


pumazi:
We'll squash this into #549, which will resolve the test build issue.

PR #554, opened 2018-04-27
Fix the docker container build to use the base python container

pumazi:
This upgrades pip by installing it as a local user installation. We no longer use the system package (python-pip) since it has an issue upgrading.

And I've made a simple adjustment in the travis config to call the generated container, whatever it's name might be rather than hard-coding to the exact name.

This corrects the failing tests on master.


pumazi:
It's essentially a move from ubuntu 16.04 to debian jessie with python preinstalled. So lighter weight and more precise IMO.

PR #556, opened 2018-05-10
Build sitemap without stripping special chars to match Webview url

mwharrison:
Most of the regx came from [here](https://stackoverflow.com/questions/4328500/how-can-i-strip-all-punctuation-from-a-string-in-javascript-using-regex/25575009#25575009).

See https://github.com/Connexions/webview/pull/1631 for the frontend aspect of this, along with its test plan.

Test plan
======
Steps
------
1. Visit a sitemap for a publisher who has published a book with unicode characters (ie.: Katalyst whose books are in a different language, https://qa.cnx.org/sitemap-katalysteducation.xml)
2. Unicode (non-ascii) characters should be present in the URLs
(ie. Expect this URL https://qa.cnx.org/contents/34972a80-efe3-4f36-90d4-e0661d202e24@2/Płyny-gęstość-i-ciśnienie to be present instead of https://qa.cnx.org/contents/34972a80-efe3-4f36-90d4-e0661d202e24@2/p-yny-g-sto-i-ci-nienie

Expected result
----------------
- Unicode characters do not become stripped out from URLs in a sitemap.
- Whitespace is replaced with dashes.
- Punctuation is stripped out (ie. question marks, exclamation marks)

### DIFF
--- a/cnxarchive/tests/data/sitemap-OpenStaxCollege.xml
+++ b/cnxarchive/tests/data/sitemap-OpenStaxCollege.xml
@@ -1,67 +1,67 @@
 <?xml version='1.0' encoding='utf-8'?>
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
   <url>
-    <loc>http://cnx.org/contents/a733d0d2-de9b-43f9-8aa9-f0895036899e@1.1/span-style-color-red-derived-span-copy-of-college-i-physics-i</loc>
+    <loc>http://cnx.org/contents/a733d0d2-de9b-43f9-8aa9-f0895036899e@1.1/-span-style-color-red-Derived-span-Copy-of-College-i-Physics-i-</loc>
     <lastmod>2013-08-31</lastmod>
   </url>
   <url>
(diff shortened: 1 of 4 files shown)


mwharrison:
[CHANGE REQUEST: allow non-ascii in URLs](https://trello.com/c/1xB3R6xO/483-change-request-allow-non-ascii-in-urls)

pumazi:
@mwharrison, I've adjusted your commits slightly, because I was going to fix the tests. But then I discovered this might not be quite what we wanted.

I believe we'll likely want to leave `-`'s in place for a kind of human readable url. I could be wrong, so check with folks about the intention here.

mwharrison:
@pumazi - that's fine, I was going to rebase before I removed the WIP (which I just realized I didn't have the label on here).
I agree - we want to leave the `-`'s, so working on that. :)

mwharrison:
@reedstrm @pumazi - I'm having issues getting this to work. If I do the query in the database, it works fine - but the tests seem to be failing on special characters. Any ideas?

screen shot 2018-05-31 at 2 34 35 pm

Dantemss:
See comments on https://github.com/Connexions/webview/pull/1631 about the difference in behavior between the 2 PRs. I also think the new test expectations seem odd (why do some urls end in `-` now?)

PR #559, opened 2018-05-31
Sitemap order by author

CecileJiaqiWang:
Segment sitemap by author.


pumazi:
There is a `.swp` file in this BTW. 😄

PR #561, opened 2018-06-26
add safe_stat util, and use it to test exports dirs before access

reedstrm:
This works around the stall that happens when file IO attempts to use a stale NFS mount. The metadata about exports that is collected here is non-critical to content serving, this allows the process to report `missing` rather than freezing

### DIFF
--- a/cnxarchive/utils/safe.py
+++ b/cnxarchive/utils/safe.py
@@ -6,24 +6,23 @@ from logging import getLogger
 
 logger = getLogger('safestat')
 
-process = None
+safe_stat_process = None
 
 
 def safe_stat(path, timeout=1, cmd=None):
     "Use threads and a subproc to bodge a timeout on top of filesystem access"
-    global process
+    global safe_stat_process
 
     if cmd is None:
         cmd = ['/usr/bin/stat']
 
     cmd.append(path)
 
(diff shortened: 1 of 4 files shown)


reedstrm:
Implementation seems a bit clunky (threads and subprocess, oh my!) but all the non-blocking/async sorts of things that I tested failed to not block on an actual stale NFS mount. Also, some of those are python3 only. This is 2&3, and fairly robust.

cnxarchive/utils/safe.py
```diff
@@ -0,0 +1,37 @@
...
+from logging import getLogger
+
+logger = getLogger('safestat')
+
+process = None
+
+
+def safe_stat(path, timeout=1, cmd=None):
+    "Use threads and a subproc to bodge a timeout on top of filesystem access"
+    global process
```
Dantemss: Since this is a global variable, it should probably have a less generic name.

cnxarchive/utils/safe.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if cmd is None:
+        cmd = ['/usr/bin/stat']
+
+    cmd.append(path)
+
+    # import pdb; pdb.set_trace()
```
Dantemss: Might want to remove this comment :)

PR #565, opened 2018-07-16
Allow the extras API to return single fields

Dantemss:
Requires AND required for https://github.com/Connexions/webview/pull/1719

### DIFF
--- a/cnxarchive/tests/views/test_extras.py
+++ b/cnxarchive/tests/views/test_extras.py
@@ -90,8 +90,8 @@ VALUES
         pyramid_testing.tearDown()
         self.fixture.tearDown()
 
-    def assert_featured_links_match(self, extras):
-        self.assertEqual(extras['featuredLinks'], [
+    def assert_featured_match(self, extras):
+        self.assertEqual(extras['featured'], [
             {u'id': u'e79ffde3-7fb4-4af3-9ec8-df648b391597',
              u'title': u'College Physics',
              u'version': u'7.1',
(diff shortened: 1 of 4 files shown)


Dantemss:
Hmm @reedstrm do you think it's odd that we request `site-messages` but get back `messages`? Maybe we should just use the same names?

cnxarchive/tests/views/test_extras.py
```diff
@@ -89,126 +83,190 @@ def test_extras(self, cursor):
...
+            {u'id': 5, u'name': u'Science and Technology',
+             u'count': {u'module': 6, u'collection': 1},
+             },
+            {u'id': 6, u'name': u'Social Sciences',
+             u'count': {u'module': 0, u'collection': 0},
+             }
+            ])
+
+    def assert_languages_and_count_match(self, extras):
+        self.assertEqual(extras['languages_and_count'], [[u'da', 1], [u'en', 19]])
```
Dantemss: This test is flaky: the order is reversed on PG on OS X.

cnxarchive/tests/views/test_extras.py
```diff
@@ -89,126 +83,190 @@ def test_extras(self, cursor):
...
+            {u'id': 5, u'name': u'Science and Technology',
+             u'count': {u'module': 6, u'collection': 1},
+             },
+            {u'id': 6, u'name': u'Social Sciences',
+             u'count': {u'module': 0, u'collection': 0},
+             }
+            ])
+
+    def assert_languages_and_count_match(self, extras):
+        self.assertCountEqual(extras['languages_and_count'], [[u'da', 1], [u'en', 19]])
```
Dantemss: This should ensure the lists have the same elements and the same counts of each element, in any order. This query does not guarantee the order of the languages.

cnxarchive/tests/views/test_extras.py
```diff
@@ -89,126 +83,190 @@ def test_extras(self, cursor):
...
+            {u'id': 5, u'name': u'Science and Technology',
+             u'count': {u'module': 6, u'collection': 1},
+             },
+            {u'id': 6, u'name': u'Social Sciences',
+             u'count': {u'module': 0, u'collection': 0},
+             }
+            ])
+
+    def assert_languages_and_count_match(self, extras):
+        self.assertEqual(set(extras['languages_and_count']), set([[u'da', 1], [u'en', 19]]))
```
Dantemss: This should ensure the lists have the same elements, in any order. This query does not guarantee the order of the languages (test fails on PG on OS X).

PR #566, opened 2018-07-19
Allow contextual urls in the page extras

Dantemss:
Should cut the time to load extras for physics pages by about half once webview uses this new route (not as much for other books since physics is probably the worst case for this issue).


Dantemss:
The redirect code for shortId is kind of a mess (the default behavior for contextual pages in extras with my changes was to loop forever, and the fix for that is duplicated in 3-4 places). Should I attempt to clean it up?

Dantemss:
One option to clean up the redirect code would be to change the `split_ident_hash` function to take the args dict and a key (that contains the ident_hash). We could then propagate the key to the error handler for the `IdentHashShortId` exception (`ident_hash_short_id`) and make that method set the correct key in the route_args.

PR #569, opened 2018-07-31
support filters

CecileJiaqiWang:
Support all the filters in advanced search.

Depends on Connexions/cnx-db#132

### DIFF
--- a/cnxarchive/search.py
+++ b/cnxarchive/search.py
@@ -516,8 +516,8 @@ def _build_search(structured_query):
             if value[0] != 'book' and value[0] != 'collection' and \
                     value[0] != 'page' and value[0] != 'module':
                 invalid_filters.append(idx)
-            value[0] = 'Collection' if (value[0] == 'book' or \
-                value[0] == 'collection') else 'Module'
+            value[0] = 'Collection' if (value[0] == 'book' or
+                        value[0] == 'collection') else 'Module'
             arguments.update({'type': value[0]})
         elif keyword == 'keyword':
             value = _upper(value)
--- a/cnxarchive/search.py
+++ b/cnxarchive/search.py
@@ -44,7 +44,7 @@ MAX_VALUES_FOR_KEYWORDS = 100
 MAX_VALUES_FOR_AUTHORS = 100
 SORT_VALUES_MAPPING = {
     'pubdate': 'revised DESC',
-     'version': 'version DESC',
+    'version': 'version DESC',
     'popularity': 'rank DESC NULLS LAST',
     }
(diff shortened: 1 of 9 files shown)


CecileJiaqiWang:
[Weighted Search](https://trello.com/c/qNXCjqec/529-weighted-search)

reedstrm:
:skull_and_crossbones:

pumazi:
I'm having issues dumping this...

```
> pg_dump ...
...
pg_dump: [archiver (db)] query failed: ERROR:  function short_id(uuid) does not exist
LINE 1:  select short_id(uuid) || '@' || concat_ws('.', major, minor...
                ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:   select short_id(uuid) || '@' || concat_ws('.', major, minor) 
CONTEXT:  SQL function "short_ident_hash" during inlining
pg_dump: [archiver (db)] query was: DECLARE _pg_dump_cursor CURSOR FOR SELECT * FROM ONLY public.modules
```

pumazi:
O... the other issue with the committed data dump file is that it is empty. 😝

pumazi:
I tried doing the `pg_dump` on 'dev' as well. I get the same error. So something systemic is happening. I'm at a lost for now, but I'll keep kicking at it.

pumazi:
I tried loading and then dumping without changing anything. This still fails at the same error.

pumazi:
It'd kinda be nice to exclude the functions from the dump, but there is no option for that. Nearest solution I found for that is to put the functions in a separate schema and exclude that schema. But that's the long way around the issue.

cnxarchive/search.py
```diff
@@ -539,81 +415,132 @@ def _build_search(structured_query, weights):
...
+    arguments.update({'fulltext_key':  ';--;'.join(fulltext_key)})
+
+    idx = 0
+    invalid_filters = []
+    filters = _convert(structured_query.filters, [])
+
+    while idx < len(filters):
+        keyword = filters[idx][0]
+        value = filters[idx][1]
+        if keyword == 'pubYear':
```
reedstrm: this long if/elif tree isn't ideal, but is at least clear as to what's going on. Note that several branches of the tree do slightly different things …

reedstrm:
ok found it - something to do with building a functional index via calling an additional custom function:
the `modules_short_ident_hash` index called `short_ident_hash`, which is defined to call `short_id`. I noticed that pg_dump likes to mess with the `search_path` (sets it to the empty string), so I redefined `short_ident_hash` to call `public.shortid`, and it suddenly can dump it.

PR #575, opened 2018-08-01
Faster book extras and more accurate contextual page extras

Dantemss:
Webview loads book extras when loading books, but never displays the "list of books this book is in". We now return an empty array for that, skipping the DB query.

For contextual page extras requests, we now return the book that was requested. Still fast, but slightly more accurate. However, we still do not check that the page is actually in the requested book.

### DIFF
--- a/cnxarchive/tests/views/test_content.py
+++ b/cnxarchive/tests/views/test_content.py
@@ -1123,7 +1123,29 @@ INSERT INTO trees (nodeid, parent_id, title, childorder, is_collated)
         with self.assertRaises(IdentHashShortId) as raiser:
             get_extra(self.request)
 
-    def test_get_extra_contextual_url(self):
+    def test_get_extra_book(self):
+        book_id = 'e79ffde3-7fb4-4af3-9ec8-df648b391597'
+        book_version = '7.1'
+
+        # Build the request
+        self.request.matchdict = {
+            'ident_hash': '{}@{}'.format(book_id, book_version),
+            'page_ident_hash': '',
+            'separator': ':'}
+        self.request.matched_route = mock.Mock()
+        self.request.matched_route.name = 'content-extras'
+
+        from ...views.content import get_extra
+        output = get_extra(self.request).json_body
+
+        self.assertEqual(self.request.response.status, '200 OK')
+        self.assertEqual(self.request.response.content_type,
+                         'application/json')
+        self.assertEqual(set(output.keys()), set([
+            'canPublish', 'downloads', 'headVersion',
+            'isLatest', 'latestVersion', 'state']))
+
+    def test_get_extra_page_contextual_url(self):
         book_id = 'e79ffde3-7fb4-4af3-9ec8-df648b391597'
         book_version = '7.1'
         page_id = '209deb1f-1a46-4369-9e0d-18674cf58a3e'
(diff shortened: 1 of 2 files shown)


cnxarchive/tests/views/test_content.py
```diff
@@ -872,7 +872,6 @@ def test_get_extra_no_allowable_types(self):
                          'application/json')
         output['canPublish'].sort()
         self.assertEqual(output, {
-            u'books': [],
```
pumazi: Shouldn't the key be available on each request so that the schema expectations are met? That is, there is a general structure, even if it has null values. Or maybe you are keying in on this a way to decipher that it is not contextual?

Dantemss: That would make sense. Would you prefer to keep the empty arrays or use nulls in this case when the query is not even performed?

pumazi: Your call on empty array or null.

TBH I'm casually thinking about this in terms of the swagger schema, which likes a complete structure to be a valid response. Or at least that's what I've experienced.

Dantemss: Asked Ross and he would rather we returned the book itself in both cases (contextual page extras and book extras). I'll work on that.

Dantemss: Hm maybe book for page and empty array for book would make more sense actually.

cnxarchive/tests/views/test_content.py
```diff
@@ -1166,6 +1244,7 @@ def test_extra_w_utf8_characters(self):
         self.assertEqual(self.request.response.content_type,
                          'application/json')
         output['canPublish'].sort()
+        output['books'].sort()
```
Dantemss: This test started failing for no reason and I had to force the order on the list of books this page is in. Guessing it was always flaky, since the list of books is not ordered.

Dantemss: This test started failing for no reason, with the books coming in the opposite order. The `get-books-containing-page` code in cnxdb does sort the books, but these books have the same exact authors and revised dates, so I think the test was always flaky. I started sorting the books here and made the expected values match.

Dantemss:
@reedstrm please confirm the returned values now match what you would expect.

Dantemss:
Also, if we want to check that the page is actually in the given book when a contextual request is made to extras (and maybe return a 404 if it's not), I suppose I could do that as well.

pumazi:
> Also, if we want to check that the page is actually in the given book when a contextual request is made to extras (and maybe return a 404 if it's not), I suppose I could do that as well.

If that's what we are currently doing, I'd say that's wrong. If you decide to do that implementation, please submit another PR.

Dantemss:
Well extras did not use to support contextual urls. I implemented them but never checked that the page is actually in the book. I will submit a separate PR for that.

reedstrm:
We do in fact 404 if you request a page in the context of a book that it is not in, under the `/contents` path. @pumazi  why do you think that is wrong for the `/extras` path? What do you suggest we do instead? (Note we also 404 when asked for a composite book outside the context of the book for which it is defined. That one will probably turn into a 301 once we start tracking a "canonical book" for each page, as per the SEO canonical URL story)

Dantemss:
I made https://github.com/Connexions/cnx-archive/pull/577 (based on this PR) that will check that the page is actually in the book and, if not, return a 404.

PR #578, opened 2018-08-06
Update readme, correct config file

BryanEliDimas:
Update the README.

1. Wrong `testing.ini` file location specified. This PR fixes it.
2. Mention that the tests can be run with `pytest`.
3. Mention that Docker (Community Edition, the free version) must be installed first before running the command `docker-compose`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -205,15 +205,15 @@ Running tests
    :target: https://codecov.io/gh/Connexions/cnx-archive
 
 The tests use the standard library ``unittest`` package and can therefore
-be run with minimal effort. Make a testing config, such as testing.ini,
-and set the environment variable ``TESTING_CONFIG`` to the name of that file::
+be run with minimal effort. Set the environment variable ``TESTING_CONFIG`` to the location of the testing.ini file::
 
-    export TESTING_CONFIG=testing.ini
+    export TESTING_CONFIG=cnxarchive/tests/testing.ini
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -205,16 +205,19 @@ Running tests
    :target: https://codecov.io/gh/Connexions/cnx-archive
 
 The tests use the standard library ``unittest`` package and can therefore
-be run with minimal effort. Make a testing config, such as testing.ini,
-and set the environment variable ``TESTING_CONFIG`` to the name of that file::
+be run with minimal effort. Set the environment variable ``TESTING_CONFIG`` to the location of the testing.ini file::
 
-    export TESTING_CONFIG=testing.ini
+    export TESTING_CONFIG=cnxarchive/tests/testing.ini
```
pumazi: This likely wasn't clear. You only need to make a copy of the config if you need to change the default testing settings. The step that is missing is probably a `cp cnxarchive/tests/testing.ini`.

reedstrm: How about "Set the environment variable ``TESTING_CONFIG`` to point to your testing configuration file. A default example can be found at ``cnxarchive/tests/testing.ini``, and can be used directly, or copied to another location and modified. Please do not modify in place unless you intend to change the defaults for everyone."

BryanEliDimas: @pumazi Ok, makes sense.

@reedstrm Perfect—thanks! I updated my commit.

PR #580, opened 2018-08-13
Set Cache-Control headers in archive

Dantemss:
Still need to test if the routes with multiple decorators will behave properly.


reedstrm:
@mulich  - we merging this w/o the test, then?

reedstrm:
@pumazi we merging w/o the test, then?

pumazi:
I don't think we need to test the framework is doing the right thing with this info. So no tests necessary. 👍

reedstrm:
Right. Meant to put that comment on the cnx-publishing Fallback PR.

PR #584, opened 2018-08-17
Add ETag header to archive responses

Dantemss:
This tween adds ETag headers to archive responses (as long as they are not too large).


openstaxalina:
Remaining steps:
-analyze 15 different views in archive
-generate appropriate ETags for each view

PR #585, opened 2018-08-18
test for html abstract trigger

reedstrm:
depends on https://github.com/Connexions/cnx-db/pull/147 tests will not pass until that is merged


reedstrm:
this test is no longer useful, moved the trigger, and there exist tests at that spot.

PR #589, opened 2018-08-30
add link header to PDFs via http

reedstrm:
This adds a 'LINK' header to the HTTP response, when returned from an /exports/*pdf url.


BryanEliDimas:
Code reviewed ✅

PR #591, opened 2018-09-04
remove version from canonical link

reedstrm:
The version is _not_ supposed to be part of the canonical URL (per https://docs.google.com/document/d/18cjuLutHYOY3eWuxf14lXiblfuJ8KbEdC4PN3a4yHEw)

### DIFF
--- a/cnxarchive/views/exports.py
+++ b/cnxarchive/views/exports.py
@@ -77,9 +77,11 @@ def get_export(request):
                                " filename*=UTF-8''{fname}".format(
                                        fname=encoded_filename)
     resp.body = file_content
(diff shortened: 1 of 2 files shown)


cnxarchive/views/exports.py
```diff
@@ -77,9 +77,10 @@ def get_export(request):
                                " filename*=UTF-8''{fname}".format(
                                        fname=encoded_filename)
     resp.body = file_content
+    slug_title = '-'.join(encoded_filename.split('-')[:-1])
```
Dantemss: Would be good to add a comment saying that this is removing the version from the slug.

PR #592, opened 2018-09-05
Cache exports for only 1 minute since they can be overwritten

Dantemss:
When collections get a minor version rev.


kerwinso:
@Dantemss Does this need code review?

PR #593, opened 2018-09-25
Add Cache-Control headers for redirects

Dantemss:
So 301's are cached forever and we don't rely on the varnish override for 302's.


Dantemss:
If you think it would help, I could change the default caching time back to 10 minutes.

PR #596, opened 2018-09-26
Revert "Add Cache-Control headers for redirects"

Dantemss:
Reverts Connexions/cnx-archive#593
Can't set the Cache-Control on all 301 legacy redirects because it's also catching the unversioned case.
Cache-Control on 302 redirects did not work for some reason.


Dantemss:
Going to poke a bit more instead.

PR #598, opened 2018-09-26
Reduce the number of stat calls in /extras

Dantemss:
Done in a way that minimizes the number of safe_stat calls as well
Also avoids reading the files for content extras

### DIFF
--- a/cnxarchive/views/content.py
+++ b/cnxarchive/views/content.py
@@ -30,7 +30,7 @@ from ..utils import (
 from .helpers import (
     get_uuid, get_latest_version, get_head_version, get_content_metadata
     )
-from .exports import get_export_file, ExportError
+from .exports import get_export_files
 
 HTML_WRAPPER = """\
 <html xmlns="http://www.w3.org/1999/xhtml">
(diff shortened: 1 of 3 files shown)


cnxarchive/views/content.py
```diff
@@ -211,25 +211,28 @@ def get_state(cursor, id, version):
...
-                'path': request.route_path(
-                    'export', ident_hash=join_ident_hash(id, version),
-                    type=type_name, ignore=u'/{}'.format(filename))
-                }
-        except ExportError as e:  # noqa
-            # Some other problem, skip it
-            pass
+    type_settings = request.registry.settings['_type_info']
+    type_names = [k for k, v in type_settings]
+    type_infos = [v for k, v in type_settings]
```
reedstrm: this could be written:
`type_names, type_infos = zip(*type_settings)`
(yes, zip is its own inverse function)
Bu it's such a short list, there's no real advantage.

cnxarchive/views/exports.py
```diff
@@ -85,55 +86,83 @@ def get_export(request):
...
-
-    for dir in exports_dirs:
-        if safe_stat(dir):
+
+    reachable_dirs = [dir for dir in exports_dirs if safe_stat(dir)]
+
+    # 1 result per type, in the same order as the given types
+    results = []
+
+    for type in list(types):
```
reedstrm: Hmm, why the extra wrap of `list()` around `types`, which is already a list? Harmless, but slightly jarring.

PR #600, opened 2018-09-26
Canonical book

reedstrm:
requires matching PR in cnx-db
This implements returning a rel canonical link header for all content.


Dantemss:
I restarted Travis so see if we have more luck (docker containers did not start properly).

PR #602, opened 2018-10-04
the title may include unicode, such as ® or ° or ™

reedstrm:
grr, unicode! (this PR should™, but does not contain a change to the tests to include a unicode title)


reedstrm:
Dante did a drive-by approval (via slack, phone, and in person) :-)

PR #603, opened 2018-10-09
Add a jenkinsfile for releasing the container and package

pumazi:
This builds the container and publishes it to the container registry.

This also builds the Python distribution/package and uploads it to pypi.org.


m1yag1:
I believe the twine stuff is configured in Jenkins but I could be wrong @pumazi

pumazi:
Ah, my description wasn't exactly accurate. I removed the test pypi options.

PR #614, opened 2018-12-01
Add the prefix filter to the docker app configuration file

pumazi:
This allows the application to handle X-Forwarded-* headers,
which is necessary when we use a reverse proxy for SSL termination.


pumazi:
works `staged.cnx.org`

PR #615, opened 2018-12-04
Remove module PDFs from results

edwoodward:
Skip module PDFs when processing results


karenc:
Could you remove cnxarchive/tests/data/exports/c0a76659-c311-405f-9a99-15c71af39325@5.txt?  I think it's no longer used in any tests.

cnxarchive/views/exports.py
```diff
@@ -103,6 +106,11 @@ def get_export_files(cursor, id, version, types, exports_dirs, read_file=True):
         if type not in type_info:
             raise ExportError("invalid type '{}' requested.".format(type))
         file_extension = type_info[type]['file_extension']
+
+        # skip module PDFs
+        if legacy_id.startswith('m') and file_extension == 'pdf':
+            continue
```
reedstrm: Hmm, at this point in the code, we _have_ the full metadata for the object the view is being built for: at minimum, this should test `metadata['mediaType'] == MODULE_MIMETYPE` (need some imports from `..utils` for that, see `views/content.py`)

Ideally, we would extend the exportable file-types config to include per-document-type-file-types, but given that archive will not be responsible for these files under the new pipelines, I think just testing against the type is enough.

PR #617, opened 2018-12-12
Fix the recent.rss feed

pumazi:
This fixes the invalid xml associated with this feed.

I've added some of the optional rss v2.0 tags to allow readers to derive uniqueness of the items, since many of the results have the same title (e.g. Introduction). Uniqueness can now be found by adding the version to the link. I've also added the `guid` tag to aid in that uniqueness discovery.


cnxarchive/tests/data/recent.rss
```diff
@@ -269,4 +269,4 @@
       
 
 </channel>
-</rss>
+</rss>
```
karenc: This is so strange... Why does recent.rss not have the newline? O_o  Is it better to change the test to strip whitespace instead?

pumazi: It might be, but this is working, so I'd rather not fiddle with it.

PR #625, opened 2019-01-24
Move ident hash code to cnx-db

karenc:
See openstax/cnx-db#174

- Move ident hash code to cnx-db

  In order to make cnx-db standalone, I decided to move the only
  cnx-archive imports -- join_ident_hash and split_ident_hash -- to
  cnx-db.
  
  It is still possible to import the ident hash functions from
  `cnxarchive.utils` but the code is not there anymore.

- Output container log if test fails to start on travis

  It used to be that travis would fail after the maximum time limit for
  the job when the archive container exited.  This is a bit inconvenient
  because the user might think the tests are just taking a long time.
  
  We have a loop that waits for the archive container to start before
  running the tests. If the archive container exits, we should print out
  the logs and not continue waiting for it to be up.

- Remove db build arg "gitcommithash" from .travis-compose.yml

  The db image does not take any args called "gitcommithash".  When
  building the db image, we get a warning message:
  
  ```
  [Warning] One or more build-args [gitcommithash] were not consumed
  ```
  
  Since it does not do anything, we should remove it so people don't get
  confused.

- Fix collection.xml comparison in collection minor update test

  In the collection minor update test, we are comparing the new
  collection.xml generated in the database with the one in test/data.  The
  code tries to remove the revised timestamp by doing:
  
  ```
              rev_offset = result.find('<md:revised>')
              rev_start = rev_offset + 12  # len('<md:revised>')
              rev_end = rev_start + 32  # len(timestamp)
              self.assertEqual(result[:rev_start]+result[rev_end:],
                               expected[:rev_start]+expected[rev_end:])
  ```
  
  This fails sometimes because timestamp appears to be not always 32
  characters long.  For example, one time it got truncated to
  `<md:revised>/md:revised>`.
  
  Avoid this issue by using regular expression to remove everything inside
  the `<md:revised>` tag.

- :skull_and_crossbones: DO NOT MERGE Use cnx-db remove-archive-dep branch

Please remove the last commit before merging.


pumazi:
We'll need to bump the cnx-db dependency from `2.3.0` to `3.0.0`. I'll touch that tomorrow morning if nobody has done so by then.

PR #628, opened 2019-02-06
Redid README in markdown and updated install instructions

edwoodward:


pumazi:
As a side note, we don't cover this kind of detailed installation in apps like Press, because containerization has eliminated the need for platform dependent installation instructions. These instructions are a burden to keep up to date, especially when the containers are easy to use, maintain and test. See also https://github.com/openstax/cnx-archive#install-using-docker.

karenc:
Thanks for rearranging and updating README so it's easier to read! :+1:

PR #635, opened 2019-06-24
XPath search with support for searching CNXML, HTML and Baked HTML content

pumazi:
This adds the ability to do an XPath search against CNXML, HTML and Baked HTML content. This is different from the previous implementation which only allowed for CNXML content to be searched.

---

If I had to do this over again, I'd not start here. IMFO this stand-alone API point shouldn't exist. It should have been written as `/contents/{id}/xpath.(json|html)` to take advantage of the already known context and logic that resides in that API space.

... we'll likely squash merge this, so don't mind the offhandedness of the commits.


pumazi:
I'm not particularly proud of this code. But it'll do for the short time that we need it. =D

pumazi:
I'm splitting out the encapsulated bit of this work. This PR now depends on #636

pumazi:
The SQL queries have been included in https://github.com/openstax/cnx-db/pull/182.

karenc:
When I tried to search for MathML in cnxml `/xpath.html?id=30189442-6998-4686-ac05-ed152b91b9de@24.1&q=//m:math`:

```
  File "/var/cnx/venvs/archive/src/cnx-archive/cnxarchive/views/templates/xpath.html", line 15, in top-level template code
    <li><pre>{{ match|e }}</pre></li>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 163: ordinal not in range(128)
```

Similarly when I searched for terms in html `/xpath.html?id=30189442-6998-4686-ac05-ed152b91b9de@24.1&q=//*[@data-type="term"]&type=html`:

```
  File "/var/cnx/venvs/archive/src/cnx-archive/cnxarchive/views/templates/xpath.html", line 15, in top-level template code
    <li><pre>{{ match|e }}</pre></li>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 60: ordinal not in range(128)
```

@pumazi Do you know how to fix this or do you need more information?

pumazi:
@karenc I believe I've addressed the situation. Please try again! 🙇

I used the dump book script to test this locally.

karenc:
> @karenc I believe I've addressed the situation. Please try again! bowing_man
> 
> I used the dump book script to test this locally.

@pumazi Yay!!!  It's working for me :tada:

pumazi:
💀 commit has been removed, which is causing the tests to fail. This is due the latest version of this project being included within the cnx-db container. This is resolved after this PR has been merged and released.

PR #643, opened 2020-04-10
Add requirements files main.txt and test.txt for standardization and pyup support

m1yag1:


tomjw64:
I'm curious - why does the dependency list look much longer now?

m1yag1:
@tomjw64 b/c it represents every high-level dependency and sub dependency. There were issues with a sub-dependency conflict so I made sure to pin everything.

m1yag1:
@tomjw64 Additionally, if you compare press and publishing you'll see this list mirrors their dependencies more closely.

tomjw64:
I ask because I've always found it silly that we pin things further down the dependency tree. But if it's because some incompatibility, I suppose that makes sense. Which dependency conflicted?

m1yag1:
@tomjw64 In this case it was a testing dependency. There were two things calling for beautifulsoup4 and one of them required soupsieve to be a lower version. This caused the conflict. While it may seem silly ... once you have spent enough hours troubleshooting this sort of thing you find it helps reduce time wasted fixing this crap. If python had a concept of gemfile.lock it would be less of a problem but it currently doesn't.

tomjw64:
Oh, interesting. I'm more familiar with npm/yarn package management - so I'm used to two dependencies being able to pin and use two completely separate versions of libraries. I'm surprised that's not possible with pip.

m1yag1:
There was one that was getting traction but it has since run into a number of issues. The maintainer went off the rails and it has slowly been losing momentum. It looked promising but I'd be worried to start implementing it.  https://github.com/pypa/pipenv

m1yag1:
https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/

m1yag1:
@tomjw64 can you review/approve plz? id like to get these changes merged.

PR #649, opened 2020-05-05
Update cnx-common and cnx-epub

philschatz:
Refs openstax/cnx#972 and openstax/cnx#924

### DIFF
--- a/requirements/main.txt
+++ b/requirements/main.txt
@@ -1,6 +1,6 @@
 certifi==2020.4.5.1
 chardet==3.0.4
-cnx-common==1.2.3
+cnx-common==1.3.3
 cnx-db==3.5.3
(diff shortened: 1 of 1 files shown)


requirements/main.txt
```diff
@@ -1,8 +1,8 @@
 certifi==2020.4.5.1
 chardet==3.0.4
-cnx-common==1.2.3
+cnx-common==1.3.3
 cnx-db==3.5.3
-cnx-epub==0.21.0
+cnx-epub==0.21.1
```
tomjw64, line 5: pin to lxml will also have to change to 4.4.3 because of this

philschatz: Ah thanks!

